PR #3, opened 2019-03-21
Add the String.reverse function

Dean177:
Documented here: https://package.elm-lang.org/packages/elm/core/1.0.2/String#reverse


Dean177:
Thats right [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) is creating a copy.

PR #5, opened 2019-03-21
Create the Tuple3 module

Dean177:


bs/__tests__/tablecloth_test.ml
```diff
@@ -3,7 +3,69 @@ open Jest
...
+
+    test "mapSecond" (fun () ->
+      expect (Tuple3.mapSecond ~f:sqrt ("stressed", 16., false)) |> toEqual ("stressed", 4., false)      
+    );
+
+    test "mapThird" (fun () ->
+      expect (Tuple3.mapThird ~f:not ("stressed", 16, false)) |> toEqual ("stressed", 16, true);
+    );
+
+    test "mapEach" (fun () ->
```
Dean177: I wasn't sure what to call this one. Suggestions?

pbiggar: Elm calls it each, so that makes sense: https://package.elm-lang.org/packages/Fresheyeball/elm-tuple-extra/latest/Tuple3#mapEach

Dean177:
Ah, I didn't see the extras module, I will include all the goodies from it too.

PR #6, opened 2019-03-21
Add additional Tuple2 & Tuple3 functions from ‘extras’

Dean177:
I have deviated slightly from the Elm extras package.
- Tuple2.mapBoth -> Tuple2.mapEach so the names are consistent with the same function in Tuple3
- swilll -> swirlLeft & swirl -> swirlRight (I assume that’s what the l and r stand for) to be a little more explicit
- init & tail -> head & tail. `init` doesn’t seem like a great name to me (short for initial?) as it doesn’t correspond with `tail`. `head` doesn’t seem ideal either as without looking at the signature I would guess that `head` behaved the same way as `first`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Install via npm by:
 
 Then add to your bsconfig.json file:
 
-`  "bs-dependencies" : ["tablecloth-bucklescript"]`
+`"bs-dependencies" : ["tablecloth-bucklescript"]`
 
 ### OCaml native
 
@@ -27,17 +27,18 @@ Install via opam:
 
 Then add to your dune file:
 
-`  (libraries (tablecloth-native ...))`
+`(libraries (tablecloth-native ...))`
 
 ### js_of_ocaml
 
-We have not built a js\_of\_ocaml specific version yet. However, the native version should work perfectly with js\_of\_ocaml.
+We have not built a js_of_ocaml specific version yet. However, the native version should work perfectly with js_of_ocaml.
 
 ## Usage
 
 By design, the usage of the libraries is the same on all platforms:
 
 Either open the module, and replace the builtin modules
+
 ```
 open Tablecloth
 
@@ -51,6 +52,7 @@ let () =
 ```
 
 Or use the fully qualified names:
+
 ```
 let () =
   "somestring"
@@ -67,25 +69,26 @@ upstreaming them. We have [an
 example](https://github.com/darklang/tablecloth/blob/master/examples/tc.ml)
 that you can copy from.
 
-
 ## Design of Tablecloth
 
 [Dark](https://darklang.com) uses multiple versions of OCaml on the frontend
 and backend:
+
 - Our backend is written in OCaml native, using [Jane Street Core](https://github.com/janestreet/core) as a standard
   library
 - Our frontend is written in [Bucklescript](https://bucklescript.github.io/) (dba [ReasonML](https://reasonml.github.io/))
 - Parts of our backend are shared with the frontend by compiling them using
-  js\_of\_ocaml, and running them in a web worker.
+  js_of_ocaml, and running them in a web worker.
 
 We discovered that it was impossible to share code between the Bucklescript
 frontend and the OCaml backend, as the types and standard libraries were very
 different:
+
 - Bucklescript uses camelCase by default, while most native libraries,
-  including Core and the OCaml standard library, use snake\_case.
+  including Core and the OCaml standard library, use snake_case.
 - The libraries in [Belt](https://bucklescript.github.io/bucklescript/api/index.html) have different names and function signatures than native OCaml and Base/Core.
 - Many OCaml libraries have APIs optimized for pipelast (`|>`), while Belt aims
-  for pipefirst (`|.`). 
+  for pipefirst (`|.`).
 - Core does not work with Bucklescript, while Belt is optimized for the JS
   platform.
 - Belt does not work in native OCaml, while Core is optimized for the native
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -175,4 +178,3 @@ license.
 ## Authors
 
 Written by [Dark](https://darklang.com). We may be [hiring](https://darklang.com/careers).
-
```
Dean177: Apologies to the changes for this file, I suspect this was VSCode being helpful and doing some auto-formatting once I saved.

Shall I revert the other changes (they don't appear to have any affect on how the file looks when rendered)

bs/src/tablecloth.ml
```diff
@@ -432,28 +432,46 @@ module Tuple2 = struct
   
   let second ((_, b) : 'a * 'b) : 'b = b
 
+  let map ~(f : 'b -> 'y) ((a, b) : 'a * 'b) : 'a * 'y = (a, f b)
```
pbiggar: Why does this work only on the second argument? If it's from upstream, do you know their justification? Might be worth looking at what Jane Street does for this in Base.

bs/src/tablecloth.ml
```diff
@@ -432,28 +432,46 @@ module Tuple2 = struct
...
 
   let map_first = mapFirst
 
-  let mapSecond ~(f : 'b -> 'y) ((a, b) : 'a * 'b) : 'a * 'y = (a, f b)
+  let mapSecond = map
 
   let map_second = mapSecond
 
-  let mapBoth ~(f : 'a -> 'x) ~(g : 'b -> 'y) ((a, b) : 'a * 'b) : 'x * 'y = (f a, g b)
+  let mapEach ~(f : 'a -> 'x) ~(g : 'b -> 'y) ((a, b) : 'a * 'b) : 'x * 'y = (f a, g b)
```
pbiggar: Using `~g` is clever! I wonder if it might be better as `~firstf` or `~ffirst` or `~f1`, etc?

Dean177:
I haven't included any of the sorting functions from the extras package.

Elm has a `compare` which is a union of `GT | LT | EQ` where as OCaml uses positive / zero / negative.

I personally favour the former, but it would be quite a big change / committing thing to introduce.

bs/src/tablecloth.ml
```diff
@@ -432,28 +432,46 @@ module Tuple2 = struct
...
   let create a b c = (a, b, c)
 
   let first ((a, _, _) : 'a * 'b * 'c) : 'a = a
 
-  let second ((_, b, _) : 'a * 'b *'c) : 'b = b
+  let second ((_, b, _) : 'a * 'b * 'c) : 'b = b
   
   let third ((_, _, c) : 'a * 'b * 'c) : 'c = c
 
+  let head ((a, b, _) : 'a * 'b * 'c): ('a * 'b) = (a, b)
```
pbiggar: The word `init` is a known word for the opposite of `tail`, so I'd use that. (`tail` is everything but the first, works with `head`, `init` is everything but the last, works with `last`).

bs/src/tablecloth.ml
```diff
@@ -462,13 +480,29 @@ module Tuple3 = struct
...
   let mapEach ~(f : 'a -> 'x) ~(g : 'b -> 'y) ~(h : 'c -> 'z) ((a, b, c) : 'a * 'b * 'c) : 'x * 'y * 'z = (f a, g b, h c)
 
-  let map_each  = mapEach
+  let map_each = mapEach
+
+  let mapAll ~(f: 'a -> 'b) (a1, a2, a3) = (f a1, f a2, f a3)
+
+  let map_all = mapAll
+
+  let swirlLeft ((a, b, c) : 'a * 'b * 'c) : ('c * 'a * 'b) = (c, a, b)
```
pbiggar: Good rename. I'm surprised at the signature. If it goes left, wouldn't we take the `first` element and move it to the back? `(a,b,c) -> (b,c,a)`? Is there a place that swirl comes from? If not maybe we should look at assembly and `rotate` which is a known thing at least.

README.md
pbiggar: Nah, it's fine. A nice format is good.

bs/src/tablecloth.ml
Dean177: I will un-rename it. Its not something I had come across before.

bs/src/tablecloth.ml
Dean177: I thought the same thing! They seem backwards to me and I couldn't think of a way for it to make sense.

I will have a look.

pbiggar: That's not an official or community package. Given that it hasn't had as much community review or design input as the other extra libs, no need to stick to closely to it.

bs/src/tablecloth.ml
Dean177: I considered not including map here (and also not in Tuple3).

Elm takes a lot of inspiration from Haskell here: https://stackoverflow.com/questions/13442544/why-does-the-2-tuple-functor-instance-only-apply-the-function-to-the-second-elem?noredirect=1&lq=1

The janestreet libs don't define it.

I am in favour of removing it

pbiggar: 👍

bs/src/tablecloth.ml
Dean177: Or maybe `~first` and `~second` to match the accessor functions?

pbiggar: My local functional guru says `~f` and `~g` and `~h` is a known idiom, so might be better to do that than invent a new scheme? I'm mostly open to anything though.

Dean177: Ok, lets leave it with `f`, `g` &`h` then.

pbiggar:
Is this ready for a rereview?

Dean177:
Yes, I think your comments have been addressed now.

pbiggar:
Wonderful, thanks again!

PR #9, opened 2019-03-23
Use belt when possible

j-m-hoffmann:
Hi there, i like your initiative. Hope you can use my changes.

### DIFF
--- a/bs/__tests__/tablecloth_test.ml
+++ b/bs/__tests__/tablecloth_test.ml
@@ -4,7 +4,11 @@ open Expect
 
 let () =
   describe "List" (fun () ->
-    test "indexedMap empty" (fun () -> expect (List.indexedMap ~f:(fun i _ -> i) []) |> toEqual []);
+    test "reverse empty list" (fun () -> expect (List.reverse []) |> toEqual []);
+    test "reverse one element" (fun () -> expect (List.reverse [0]) |> toEqual [0]);
+    test "reverse two elements" (fun () -> expect (List.reverse [0;1]) |> toEqual [1;0]);
+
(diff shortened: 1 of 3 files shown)


j-m-hoffmann:
i allready tried, but have not figured out alcotest yet

pbiggar:
Ah yes, it's a little complicated.

Basically, you write `AT.check` calls which are roughly equivalent to `expect` calls in bs-jest. For example:

`AT.check AT.int "length" (String.length "123") 3;`

is a test called "length" which comparsed `String.length "123"` with `3`. The complex thing is the `AT.int` which is a "testable" for the type you're trying to compare. There's a bunch of existing testables that you can compose, such as `AT.list AT.int` which makes an int list testable. 

So for example:

`test "map2 two elements" (fun () -> expect (List.map2 ~f:(+) [1;2] [1;2]) |> toEqual [2;4]);`

would be 

`AT.check (AT.list AT.int) "map2 two elements" (List.map2 ~f:(+) [1;2] [1;2]) [2;4]`

You can also create your own testables with a pretty printer and an equality function, for example:

```
let trio a b c = 
  let eq (a1, b1, c1) (a2, b2, c2) = AT.equal a a1 a2 && AT.equal b b1 b2 && AT.equal c c1 c2 in
  let pp ppf (x, y, z) = Fmt.pf ppf "@[<1>(@[%a@],@ @[%a@],@ @[%a@])@]" (AT.pp a) x (AT.pp b) y (AT.pp c) z in
  AT.testable pp eq
```

Which is much more complicated than any human should have to bear, but the idea is that you create a testing with a pretty printer and an equality function, the pretty printer using the [Fmt](https://erratique.ch/software/fmt/doc/Fmt.html) module which is a bit of a beast.

j-m-hoffmann:
thanks for the explanation, allready figured it out by looking at the tuple tests

Issue #7, opened 2019-03-21
Add ocamlformat


pbiggar:
We should ocamlformat using:

```
profile = sparse
break-string-literals = never
escape-strings = preserve
```

This would be a check in CI (once it's set up) to verify that everything is properly formatted.

pbiggar:
This is getting serious. We should:

- ocamlformat every file
- check that ocamlformat works on CI

wpcarro:
I don't mind giving this a go. I'll try and PR something today.

Dean177:
@wpcarro I actually started this aaages ago: https://github.com/Dean177/tablecloth/tree/ocamlformat

But had trouble getting it working with the bucklescript side of things, that might be a good starting point. (Maybe just to check it works for the native side and submit that first)

wpcarro:
@Dean177 thanks for chiming in. I'm testing out a configuration that I'm *hoping* should work here: #70. I'm expected the CI integration to fail at first, but I wanted to get something up to get some feedback.

Would love your review if you don't mind! Might not be worth looking at until I sort out the build issue. Should hopefully resolve that sooner than later.

wpcarro:
Should we close this now because of #70 ?

PR #12, opened 2019-03-23
Use belt when possible

j-m-hoffmann:


bs/__tests__/tablecloth_test.ml
```diff
@@ -17,7 +17,9 @@ let () =
     test "indexedMap two elements" (fun () -> expect (List.indexedMap ~f:(fun i _ -> i) ['a';'b']) |> toEqual [0;1]);
 
     test "init empty list" (fun () -> expect (List.init []) |> toEqual None);
+    (* This seems kinda wrong, if this is correct, delete the comments, proposed solution in tablecloth.ml *)
     test "init one element" (fun () -> expect (List.init ['a']) |> toEqual (Some []));
```
pbiggar: `List.init` is like `List.tail`, so whatever List.tail does, List.init should do too.

`List.tail []` => `None`
`List.tail [1]` => []`
`List.tail [1;2]` => [2]`

`List.init []` => `None`
`List.init [1]` => []`
`List.init [1;2]` => [1]`

bs/src/tablecloth.ml
```diff
@@ -276,8 +276,7 @@ module List = struct
 
   let insert_at = insertAt
 
-  let splitWhen ~(f : 'a -> bool) (l : 'a list) : ('a list * 'a list) option
-      =
+  let splitWhen ~(f : 'a -> bool) (l : 'a list) : ('a list * 'a list) option =
```
pbiggar: For the styling, I'm using ocamlformat with these settings: https://github.com/darklang/tablecloth/issues/7

We should standardize on them at some point, but until then any format is fine.

bs/__tests__/tablecloth_test.ml
j-m-hoffmann: thats what i thought so there was a bug

j-m-hoffmann: ah no, everything worked as intended

PR #14, opened 2019-03-24
Add Array module

j-m-hoffmann:


pbiggar:
Thanks for all this. Let me know when it's ready for a review.

Can you also speak a little about the goals behind the module, how you think about the stuff you're building here, and the value behind it for users? Just want to make sure we're all on the same page as I review this. Thanks!

j-m-hoffmann:
I justed started by porting the List module.

Users can use exactly the same functions on lists and arrays when it is done

pbiggar:
Great, thanks! I'd love to get this merged, so it might be useful to stop adding new things and just getting it in a shape to be merged?

I had a quick question about what modules were exposed: does this expose a new TC_Array module? I'm not sure that's desirable (I'm assuming that's a development artifact)?

j-m-hoffmann:
Yeah good idea.

i thought in the end tablecloth.ml should look like belt/base.ml with the modules split up into their own file.

module Array = Tablecloth_Array
module List = Tablecloth_List
etc

j-m-hoffmann:
i don't know yet when i will find the time

PR #15, opened 2019-03-25
Add a .circleci/config.yml

### LINKED ISSUE #8: Fix CircleCI
I hooked up tablecloth to CircleCI, but I didn't make any effort to make it work.

There's two test scripts that could be run. It might be necessary to use `esy` or similar to have a repeatable install for opam packages.

Dean177:
This PR adds a config.yml which as separate steps in a workflow builds and tests both the Bucklescript and native packages.

Closes #8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
             - v1-native-dependencies-
       # m4 is a system dependency required by conf-m4 -> ocamlfind -> fmt -> alcotest
       - run: sudo apt-get install -y m4
-      - run: opam install alcotest base dune junit junit_alcotest -y
+      - run: opam update && opam install alcotest base dune junit junit_alcotest -y && eval $(opam env)
       - run: dune build
       - run: dune runtest -f
       - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,9 +42,8 @@ jobs:
       # m4 is a system dependency required by conf-m4 -> ocamlfind -> fmt -> alcotest
       - run: sudo apt-get install -y m4
       - run: opam update && opam install alcotest base dune junit junit_alcotest -y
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+      - checkout:
+          path: ~/repo
+      - restore_cache:
+          keys:
+            - v1-native-dependencies-{{ .Branch }}-{{ checksum "tablecloth-native.opam" }}
+            - v1-native-dependencies-{{ .Branch }}-
+            - v1-native-dependencies-
+      # m4 is a system dependency required by conf-m4 -> ocamlfind -> fmt -> alcotest
+      - run: sudo apt-get install -y m4
+      - run: opam update && opam install alcotest base dune junit junit_alcotest -y
```
pbiggar: I was just reading that opam2 has a flag for this: `opam init --enable-shell-hook`, which should do make `opam config exec` unnecessary.

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+      - run: opam config exec -- dune build
+      - run: opam config exec -- dune runtest -f
+      - save_cache:
+          key: v1-native-dependencies-{{ .Branch }}-{{ checksum "native/tablecloth-native.opam" }}
```
pbiggar: Looks like saving the cache doesnt work: https://circleci.com/gh/Dean177/tablecloth/13

.circleci/config.yml
Dean177: I tried it here: 
https://github.com/darklang/tablecloth/pull/15/commits/711285c736c19332040fced1dbef66b8daeed6bb

Without any success. Any ideas what I am doing wrong?

pbiggar: My guess is I misunderstood what that does. Let's just stick with the reverted version.

PR #16, opened 2019-03-26
Possible format for documentation.

jdeisenberg:
Suggested format for documentation. Part of `List` module is documented.


documentation/README.md
```diff
@@ -0,0 +1,32 @@
...
+```reason
+let square: int => int;
+```
+
+Description of function goes here. To mark inline code as OCaml,
+follow it with {:.ocaml}; to mark it as ReasonML, follow it
+with {:.reason}. (This is a convention from kramdown
+https://kramdown.gettalong.org/) Example:
+
+`square n`{:.ocaml} `square(n)`{:.reason} returns the square of `n`.
```
pbiggar: What's the goal of this?

It doesn't seem to work in the github preview, fyi.

pbiggar:
Thanks for this!

What's the plan from here? My major motivation is to get the mli's documented, and then secondarily to have a page on the internet with the docs on it. I can see how this leads to the second, but not the first. Wdyt?

documentation/README.md
jdeisenberg: The goal is to be able to generate OCaml and Reason documentation by having a program that extracts the relevant information, or to generate an HTML page with some JavaScript magic that lets you display the code in your desired language. (Github pages does recognize the attributes, according to https://stackoverflow.com/questions/23226224/inline-code-syntax-highlighting-in-github-markdown)

pbiggar: Gotcha. Now that I see how this is going, I think we should change strategies. Let's just get the mlis documented, and worry about html later.

jdeisenberg: OK; then does that mean the mli comments should be in ocamldoc format? If so, that leaves ReasonML users out in the cold.

pbiggar: They don't need to be in ocamldoc format either. The format doesn't matter. The important thing is getting content into the mli files that explain what the functions do. (And that show up in editors via merlin, vscode lsp, etc - but this can be a separate step).

Once we're in a good shape there, we can start thinking about how to reformat for best reuse, etc.

jdeisenberg: Oh, OK. Then if you have no objection , I will keep this markdown format in the `.mli` file, unless you have a better suggestion for handling both ReasonML and OCaml.

pbiggar: Sure, sounds good. Thanks!

bs/src/tablecloth.mli
```diff
@@ -1,68 +1,476 @@
+(** Documentation for tablecloth.mli *)
```
pbiggar: Don't need this line anymore.

bs/src/tablecloth.mli
```diff
@@ -1,68 +1,476 @@
...
+      price *. (1.0 -. (percentage /. 100.0));
+      
+    map2(~f=discount, [100.0, 85.0, 30.0], [10.0, 20.0, 30.0]) ==
+      [90.0, 68.0, 21.0]
+    ```
+  *)
   val map2 : f:('a -> 'b -> 'c) -> 'a list -> 'b list -> 'c list
 
+  (**
+    `getBy ~f: predicate xs`{:.ocaml} `getBy(~f=predicate, xs)`{:.reason} returns `Some value`{:.ocaml} `Some(value)`{:.reason} for the first value in `xs` that satisifies the `predicate` function; returns `None` if no element satisifies the function.
```
pbiggar: There's an extra space after `~f:`

bs/src/tablecloth.mli
```diff
@@ -1,68 +1,476 @@
...
+    ```reason
+    let even = (x: int): bool => {x mod 2 == 0};
+    getBy(~f=even, [1, 4, 3, 2]) == Some(4);
+    getBy(~f=even, [15, 13, 11]) == None;
+    ```
+  *)
   val getBy : f:('a -> bool) -> 'a list -> 'a option
 
+  (**
+    Same as getBy
```
pbiggar: It seems that sometimes you put the doc on the underscore function, and sometimes on the camelCase version. Is there a heuristic to choose them?

bs/src/tablecloth.mli
jdeisenberg: According to http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec351, “The first special comment of the file is the comment associated  with the whole module.“ Without this comment, the first comment for the `<|` operator would become the module comment.

bs/src/tablecloth.mli
jdeisenberg: I went by whatever came first in the file. As to separating the ReasonML and OCaml comments, what do I do for functions like `map` or `any`, which are the same for both?

I have no idea how to rebase out the deleted files. My git workflow usually involves deleting everything, re-cloning, and starting all over.

pbiggar: I guess for `map` and `any`, keep them combined? I think the main thing I want to avoid is the separate reason and ocaml tags which makes it very hard to read in the mli. Perhaps "`square n` (`square(n)` in reason)` returns the square of `n`" might be more readable for now?

bs/src/tablecloth.mli
pbiggar: Ah, gotcha!

bs/src/tablecloth.mli
pbiggar: I can do the rebase before merging.

pbiggar:
(Note to self: need to rebase out the documentation dir before merging!)

bs/src/tablecloth.mli
jdeisenberg: For functions like `map` and `any`, I'll use your wording in the description and keep the Ocaml and Reason examples together as they currently are. For something like `get_by`, the description and example will be OCaml only; for `getBy`, the description and example will be ReasonML only. Both of them will have a “Same as <other function>“ note in the description. Does that work for you?

pbiggar: Yes, wonderful! And apologies for the back and forth on this - excited to get this merged :)

jdeisenberg: Not a problem, I will fix what I have so far and then commit it so you can see if we are on track before going through all 200+ functions :)

pbiggar: Fantastic, thanks again!

pbiggar:
Is this ready for a rereview?

jdeisenberg:
I have not finished all the functions, but yes, please take a look at what I have so far.

pbiggar:
@jdeisenberg Great. Took a look, looks good to merge.

pbiggar:
Thanks again for this! Hopefully things will be smooth sailing from here :)

jdeisenberg:
I'm not sure how to update this further - with a new pull request?

pbiggar:
> I'm not sure how to update this further - with a new pull request?

Yes, exactly!

jdeisenberg:
I have written a few more functions, and have broken up the long lines of text into separate lines, much as I have seen in the documentation for `Belt`.

PR #17, opened 2019-03-26
Add the Char module

Dean177:


pbiggar:
Wonderful thanks! I'll take a closer look later today I expect.

Dean177:
Sorry, I intended to include this in the PR description.

It differs from Elm's Char by trying to adhere a little more closely to [Elms design guidelines](https://package.elm-lang.org/help/design-guidelines#use-human-readable-names): 
- `toUpper` -> `uppercase` (taken from `Base`)
- `toLower` -> `lowercase` (taken from `Base`)
- `isUpper` -> `isUppercase`
- `isLower` -> `isLowercase`
- `isAlpha` -> isAlphabetic`
- `isAlphaNum` -> `isAlphanumeric`
- `isDigit` -> `isNumeric`
- `isOctDigit` / `isHexDigit` -> Not yet implemented
- `fromCode` returns an `option` rather than defaulting to a special character.

And it differs from `Base.Char` by:
`is_print` -> `isPrintable`
`of_string` -> `fromString` and it now returns an `option`
`escaped` -> Not yet implemented
`get_digit` -> Not yet implemented

pbiggar:
Gotcha. 

> `toUpper -> uppercase (taken from Base)`

Why did you drop the `to` in `toUppercase`?

> isDigit -> isNumeric

Might be worthwhile having isDigit being a synonym of isNumeric - wdyt?

> isAlpha -> isAlphabetic

Alphabetic isn't a common word. I wonder if `isLetter` might be a better name for this? Of course, the extended ASCII chars include things like á which are letters I guess. This points at `isRomanLetter` which isn't great. `isAlphabetical`? Any thoughts?

native/src/tablecloth.ml
```diff
@@ -490,13 +490,56 @@ module Result = struct
...
+  let fromCode (i : int) : char option = 
+    if 0 <= i && i <= 255 then Some (Char.chr i) else None
 
   let from_code = fromCode
+
+  let toString = Base.Char.to_string
+
+  let to_string = toString
+
+  let fromString (str : string) : char option = match String.length str with
```
pbiggar: Can you add a doc describing this?

native/src/tablecloth.ml
```diff
@@ -490,13 +490,56 @@ module Result = struct
...
 
 module Char = struct
-  let toCode (c : char) : int = Char.code c
+  let toCode (c : char) : int = Base.Char.to_int c
 
   let to_code = toCode
 
-  let fromCode (i : int) : char = Char.chr i
+  let fromCode (i : int) : char option = 
+    if 0 <= i && i <= 255 then Some (Char.chr i) else None
```
pbiggar: Can you add a doc describing this?

native/src/tablecloth.mli
```diff
@@ -220,13 +220,189 @@ module Option : sig
...
+
+    [isLower 'z' = true]
+
+    [isLower '0' = false]
+
+    [isLower 'A' = false]
+
+    [isLower '-' = false]
+
+    [isLower 'ã' = false] *)
```
pbiggar: Good example.

native/test/test.ml
```diff
@@ -1,6 +1,49 @@
...
+
+  AT.check AT.string "toString" (Char.toString 'a') "a";
+
+  AT.check (AT.option AT.char) "fromString - one-length string return Some" (Char.fromString "a") (Some 'a');
+  AT.check (AT.option AT.char) "fromString - multi character strings return none" (Char.fromString "abc") None;
+  AT.check (AT.option AT.char) "fromString - zero length strings return none" (Char.fromString "") None;
+
+  AT.check AT.char "lowercase - converts uppercase ASCII characters to lowercase" (Char.lowercase 'A') 'a';
+  AT.check AT.char "lowercase - perserves lowercase characters" (Char.lowercase 'a') 'a';
+  AT.check AT.char "lowercase - perserves non-alphabet characters" (Char.lowercase '7') '7';
```
pbiggar: Can you add a check for extended ascii chars?

pbiggar: And also to the other ascii functions?

pbiggar: I primarily want to check it's the same for native and bs.

pbiggar:
OK, this looks great. Some small questions and comments, should be very close to merging :)

Dean177:
> Why did you drop the to in toUppercase?

Just to be consistent with `Base`. I think `toUppercase` is a better name actually, since its more consistent with the other `to...` functions. I will change it.

> Might be worthwhile having isDigit being a synonym of isNumeric.
> Alphabetic isn't a common word. I wonder if isLetter might be a better name for this? Of course, the extended ASCII chars include things like á which are letters I guess. This points at isRomanLetter which isn't great. isAlphabetical? Any thoughts?

I was a little torn about these names.

I chose the current ones since it they are consistent with each other (alphabetic + numeric = alphanumeric), but I think that's a great reason to keep them 

I actually briefly named `isAlphabetic` `isLetter`, and now you have mentioned it too it has cemented in my mind as definitely a better name.

The docstring does mention that `isLetter` only detects ASCII letters but maybe that isn't enough?  Some of the functions in the ocaml standard library now have an `_ascii` suffix to indicate they only operate on those characters, perhaps that is an approach worth considering.

I think `isDigit` is probably better than `isNumeric` and will change that too (and add the `toDigit` function 😉).

pbiggar:
`isASCIILetter`? Not sure I like it.

Dean177:
Me either

pbiggar:
OK so `isLetter` with a note in the docstring?

pbiggar:
@Dean177 Is this ready?

Dean177:
Yes it is.

pbiggar:
Nice job, thanks!

PR #20, opened 2019-03-28
Cache the correct directory for native dependencies

### LINKED ISSUE #19: Cache on native CI broken
Every native build on CI reinstalls the opam components, see for example https://circleci.com/gh/darklang/tablecloth/21.

The cache should be storing them. Not sure why it isn't - perhaps opam is storing things in a different directory than we're caching?

Dean177:
Resolves #19

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,16 +36,16 @@ jobs:
           path: ~/repo
       - restore_cache:
           keys:
-            - v1-native-dependencies-{{ .Branch }}-{{ checksum "tablecloth-native.opam" }}
-            - v1-native-dependencies-{{ .Branch }}-
-            - v1-native-dependencies-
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - save_cache:
           key: v1-native-dependencies-{{ .Branch }}-{{ checksum "tablecloth-native.opam" }}
           paths:
-            - ~/opam
+            - ~/.opam
```
pbiggar: I think you'll need to bump the version to v2 aswell - presumably the fallback cache already exists and we need to break it.

Dean177:
The cache key has been updated for native dependencies so this is ready to be merged.

pbiggar:
Wonderful, thanks!

PR #21, opened 2019-03-30
Add minimum function to List.

jdeisenberg:
The functions `maximum_by` and `maximum` exist, but only `minimum_by`. For symmetry, add `minimum` function.


pbiggar:
Great, thanks! As a followup, a couple of tests or a docstring would be really nice! 😃

PR #25, opened 2019-03-31
Make splitWhen consistent with splitAt.

jdeisenberg:
Is there a compelling reason for `splitWhen` to return an `option` value? 

Right now, when there is no item satisfying the predicate in `splitWhen`, it returns `None`. This pull request suggests instead returning a non-option tuple with an empty list as its second item. This makes `splitWhen`’s behavior consistent with `splitAt` when given an `~index` equal to the list length.

```
let even x = (x mod 2 == 0)
splitWhen ~f:even  [5; 1; 2; 6; 3] = ([5; 1], [2; 6; 3])
splitWhen ~f:even  [2; 6; 5] = ([], [2; 6; 5])
splitWhen ~f:even [5; 1; 3] = ([5; 1; 3], [])
splitWhen ~f:even [] = ([], [])
```

### DIFF
--- a/bs/__tests__/tablecloth_test.ml
+++ b/bs/__tests__/tablecloth_test.ml
@@ -104,6 +104,13 @@ let () =
       test "partition one element" (fun () -> expect (List.partition ~f:(fun x -> x mod 2 = 0) [1]) |> toEqual ([], [1]));
       test "partition four elements" (fun () -> expect (List.partition ~f:(fun x -> x mod 2 = 0) [1;2;3;4]) |> toEqual ([2;4], [1;3]));
     );
+ 
+    describe "split_when" (fun () ->
+      test "split_when four elements" (fun () -> expect (List.split_when ~f:(fun x -> x mod 2 = 0) [1;3;2;4]) |> toEqual ([1;3], [2;4]));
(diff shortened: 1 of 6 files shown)


pbiggar:
Looks good, thanks! I rebased and fixed the merge conflict.

Dean177:
Isn’t this now the same function as ‘partition’?

pbiggar:
Good point. Which does Elm use - we should do that?

jdeisenberg:
Much to my amazement, Elm uses the original version of `splitWhen` (returning an `option` value). I am not sure why going into `option`-land adds anything to the party.

Dean177:
Where are you seeing that @jdeisenberg? I can only find [List.partition](https://package.elm-lang.org/packages/elm/core/latest/List#partition).

jdeisenberg:
@Dean177 I found it here: https://package.elm-lang.org/packages/elm-community/list-extra/6.1.0/List-Extra

jdeisenberg:
`partition` and `splitWhen` are not the same:
```
let even x = (x mod 2 == 0)
splitWhen ~f:even [5; 1; 2; 6; 3] = ([5; 1], [2; 6; 3])
partition ~f:even [5; 1; 2; 6; 3] = ([2; 6], [5; 1; 3])
```
`splitWhen` divides the list at the place where the predicate first succeeds; `partition` dvides the list into all the elements for which the predicate succeeds and all the elements for which it fails.

Issue #22, opened 2019-03-31
duplicated function


j-m-hoffmann:
We have got flatten and concat which do the same.
Elm and Base use concat.
Remove flatten?

pbiggar:
Sure. Let's follow the Elm API as much as possible for now, as that will reduce how much we have to think right now.

PR #27, opened 2019-04-01
Add the Array module

Dean177:
I started from @j-m-hoffmann 's pull request #14 and in an effort to have a more manageable PR I 
- removed a lot of the functions, so it is still quite incomplete compared to `List` or `Base` / `Belt` 
- Added tests and documentation for each of the functions

I also added a teeny Int module since it made writing the examples / tests easier.

This differs from Elms Array module in the following ways:
- The addition of the `singleton`, `range`, `map2`, `flatMap`, `set` `any`, `all`, `append`, `concatenate`, `flatten`, `intersperse`, `reverse`, `reverseInPlace` & `forEach` functions.
- The omission of `push`
- `indexedMap` -> `mapWithIndex`
- `foldl` -> `foldLeft` & `foldr` -> `foldRight`, additionally both functions have slightly different signatures
- `slice` uses a named and optional parameter

### DIFF
--- a/bs/src/tablecloth.ml
+++ b/bs/src/tablecloth.ml
@@ -87,8 +87,6 @@ module Array = struct
 
   let concatenate (ars : 'a array array) : 'a array = Belt.Array.concatMany ars
 
-  let flatten = concatenate
-
   let intersperse ~(sep : 'a) (array : 'a array) : 'a array = 
     Belt.Array.makeBy
       (max 0 (length array * 2 - 1)) 
--- a/bs/src/tablecloth.ml
+++ b/bs/src/tablecloth.ml
@@ -1,7 +1,3 @@
-module Caml = struct
-  module Array = Array
-end
-
 let ( <| ) a b = a b
 
 let ( >> ) (f1 : 'a -> 'b) (f2 : 'b -> 'c) : 'a -> 'c = fun x -> x |> f1 |> f2
@@ -94,8 +90,8 @@ module Array = struct
   let flatten = concatenate
 
   let intersperse ~(sep : 'a) (array : 'a array) : 'a array = 
-    Array.init
-      (max 0 (Array.length array * 2 - 1)) 
+    Belt.Array.makeBy
+      (max 0 (length array * 2 - 1)) 
       (fun i -> 
         if i mod 2 <> 0 then sep else array.(i / 2)
       )
@@ -115,7 +111,7 @@ module Array = struct
     in    
     
     if sliceFrom >= sliceTo then empty else (
-      Array.init (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
+      Belt.Array.makeBy (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
     )
 
   let foldLeft ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
@@ -169,7 +165,7 @@ module List = struct
 
   let elemIndex ~(value : 'a) (l : 'a list) : int option =
     l
-    |> Caml.Array.of_list
+    |> Belt.List.toArray
     |> Js.Array.findIndex (( = ) value)
     |> function -1 -> None | index -> Some index
 
@@ -886,7 +882,7 @@ module StrSet = struct
 
   let to_list = toList
 
-  let ofList (s : value list) : t = s |> Caml.Array.of_list |> Set.fromArray
+  let ofList (s : value list) : t = s |> Belt.List.toArray |> Set.fromArray
 
   let of_list = ofList
 
@@ -936,7 +932,7 @@ module IntSet = struct
 
   let to_list = toList
 
-  let ofList (s : value list) : t = s |> Caml.Array.of_list |> Set.fromArray
+  let ofList (s : value list) : t = s |> Belt.List.toArray |> Set.fromArray
 
   let of_list = ofList
 
--- a/bs/src/tablecloth.ml
+++ b/bs/src/tablecloth.ml
@@ -1,3 +1,7 @@
+module Caml = struct
+  module Array = Array
+end
+
 let ( <| ) a b = a b
 
 let ( >> ) (f1 : 'a -> 'b) (f2 : 'b -> 'c) : 'a -> 'c = fun x -> x |> f1 |> f2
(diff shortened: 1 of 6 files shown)


pbiggar:
This is great thanks!

What's happening with the odoc stuff? That's suddenly very unreadable.

We're also at a place where there's like 4 of us using different ocamlformat settings, and this PR is suffering from it in particular. Could you take a glance at https://github.com/darklang/tablecloth/issues/7?

bs/src/tablecloth.mli
```diff
@@ -1,1203 +1,424 @@
...
-      price *. (1.0 -. (percentage /. 100.0));
-      
-    map2(~f=discount, [100.0, 85.0, 30.0], [10.0, 20.0, 30.0]) ==
-      [90.0, 68.0, 21.0]
-    ```
-  *)
+[@@@ocaml.text
+  " Documentation for tablecloth.mli \n\nFunction names that are all lower case have their descriptions and examples in both OCaml and ReasonML format.\n\nFunction names that are in snake_case have their documentation written in OCaml format.\n\nFunction names that are in camelCase have their documentation written in ReasonML format.\n"]
+val (<|) : ('a -> 'b) -> 'a -> 'b[@@ocaml.doc
+                                   "\n  The `<|` operator applies a function to an argument. It is equivalent to the `@@` operator, and its main use is to avoid needing extra parentheses.\n\n  ### Example\n\n  ```ocaml\n  let sqr x = x * x\n  let result = sqr |< 25 (* 625 *)\n  ```\n\n  ```reason\n  let sqr = (x) => {x * x};\n  let result = sqr |< 25  /* 625 */\n  ```\n"]
```
pbiggar: This is what I mean by the odoc stuff.

pbiggar:
> What do you mean by odoc stuff?

This stuff here: https://github.com/darklang/tablecloth/pull/27/files#diff-484204121b3d8dfb8718cebe802a55a9R2

Dean177:
I have rebased out the changes to `bs/tablecloth.mli`.

pbiggar:
Great, thanks.

So a thing that I just noticed is that this array is mutable. The Elm one is not of course. The Belt one is mutable, as are the Batteries one and the OCaml one.

I'd love to hear some thoughts on this.

bs/src/tablecloth.ml
```diff
@@ -740,7 +886,7 @@ module StrSet = struct
 
   let to_list = toList
 
-  let ofList (s : value list) : t = s |> Array.of_list |> Set.fromArray
+  let ofList (s : value list) : t = s |> Caml.Array.of_list |> Set.fromArray
```
j-m-hoffmann: Belt.List.toArray? Same below

bs/src/tablecloth.ml
```diff
@@ -36,7 +169,7 @@ module List = struct
 
   let elemIndex ~(value : 'a) (l : 'a list) : int option =
     l
-    |> Array.of_list
+    |> Caml.Array.of_list
     |> Js.Array.findIndex (( = ) value)
```
j-m-hoffmann: Belt.List.toArray

bs/src/tablecloth.ml
```diff
@@ -790,7 +936,7 @@ module IntSet = struct
 
   let to_list = toList
 
-  let ofList (s : value list) : t = s |> Array.of_list |> Set.fromArray
+  let ofList (s : value list) : t = s |> Caml.Array.of_list |> Set.fromArray
```
j-m-hoffmann: Belt.List.toArray

j-m-hoffmann:
hey Dean, good job so far. As you can see in the benchmark in PR #26, using flip to force functions from Ocaml that use accumulator first to conform to the Elm Style of element first is a bad idea™

bs/src/tablecloth.ml
```diff
@@ -6,6 +10,135 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+      else max 0 (min (length array) (length array + defaultTo))
+    in    
+    
+    if sliceFrom >= sliceTo then empty else (
+      Array.init (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
+    )
+
+  let foldLeft ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
+    Belt.Array.reduce a initial (flip f)
+
```
j-m-hoffmann: flip can have an overhead of more than 300% on lists, i only use it on native/foldr to get all functions uniform.
Think we should stay with the ocaml style of doing things: ~(f : 'b -> 'a -> 'b)

bs/src/tablecloth.ml
```diff
@@ -6,6 +10,135 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+  let all ~(f : 'a -> bool) (a : 'a array) : bool = Belt.Array.every a f
+
+  let append (a : 'a array) (a' : 'a array)  : 'a array = Belt.Array.concat a a'
+
+  let concatenate (ars : 'a array array) : 'a array = Belt.Array.concatMany ars
+
+  let flatten = concatenate
+
+  let intersperse ~(sep : 'a) (array : 'a array) : 'a array = 
+    Array.init
```
j-m-hoffmann: Array.init -> initialize

bs/src/tablecloth.ml
```diff
@@ -6,6 +10,135 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+      (max 0 (Array.length array * 2 - 1))
```
j-m-hoffmann: Array.length -> length

bs/src/tablecloth.ml
```diff
@@ -6,6 +10,135 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+      if from >= 0 then min (length array) from 
+      else max 0 (min (length array) (length array + from))
+    in    
+    let sliceTo = 
+      if defaultTo >= 0 then min (length array) defaultTo 
+      else max 0 (min (length array) (length array + defaultTo))
+    in    
+    
+    if sliceFrom >= sliceTo then empty else (
+      Array.init (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
```
j-m-hoffmann: Array.init -> initialize

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+
+  let mapWithIndex  ~(f : 'int -> 'a -> 'b) (a : 'a array) : 'b array = Base.Array.mapi a ~f
+  
+  let map_with_index = mapWithIndex
+  
+  let mapi = mapWithIndex 
+
+  let map2 ~(f : 'a -> 'b -> 'c) (a : 'a array) (b : 'b array) : 'c array =
+    let minLength = min (length a) (length b) in
+    Base.Array.init minLength ~f:(fun i -> f a.(i) b.(i))
```
j-m-hoffmann: Base.Array.init -> initialize

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+
+  let map3 ~(f : 'a -> 'b -> 'c -> 'd) (arrayA : 'a array) (arrayB : 'b array) (arrayC : 'c array) : 'd array =
+    let minLength : int = Base.Array.fold ~f:Base.min ~init:(length arrayA) [|length arrayB; length arrayC|] in
+    Base.Array.init minLength ~f:(fun i -> f arrayA.(i) arrayB.(i) arrayC.(i))
```
j-m-hoffmann: Base.Array.init -> initialize

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+      if from >= 0 then min (length array) from 
+      else max 0 (min (length array) (length array + from))
+    in    
+    let sliceTo = 
+      if defaultTo >= 0 then min (length array) defaultTo 
+      else max 0 (min (length array) (length array + defaultTo))
+    in    
+    
+    if sliceFrom >= sliceTo then empty else (
+      Array.init (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
```
j-m-hoffmann: Array.init -> initialize

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+    )
+
+  let foldLeft ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
+    Base.Array.fold ~f:(fun b a -> f a b) ~init:initial a
+
```
j-m-hoffmann: flip is bad for performance, i think we should only flip in native/foldRight to get to a uniform ~(f : 'b -> 'a -> 'b) on all folds

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+  let flat_map = flatMap
+
+  let find ~(f : 'a -> bool) (a : 'a array) : 'a option = Base.Array.find a ~f
+
+  let append (a : 'a array) (a' : 'a array)  : 'a array = Base.Array.append a a'
+
+  let concatenate  (al : 'a array array) : 'a array = Base.Array.concat (Base.Array.to_list al)
+
+  let flatten = concatenate
+
```
j-m-hoffmann: you can remove flatten

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+    
+    if sliceFrom >= sliceTo then empty else (
+      Array.init (sliceTo - sliceFrom) (fun i -> array.(i + sliceFrom))
+    )
+
+  let foldLeft ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
+    Base.Array.fold ~f:(fun b a -> f a b) ~init:initial a
+
+  let fold_left = foldLeft
+
```
j-m-hoffmann: add a foldl alias?

native/src/tablecloth.ml
```diff
@@ -6,6 +6,133 @@ let ( << ) (f1 : 'b -> 'c) (f2 : 'a -> 'b) : 'a -> 'c = fun x -> x |> f2 |> f1
...
+  let foldLeft ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
+    Base.Array.fold ~f:(fun b a -> f a b) ~init:initial a
+
+  let fold_left = foldLeft
+  
+  let foldRight ~(f : 'a -> 'b -> 'b) ~(initial : 'b) (a : 'a array) : 'b =
+    Base.Array.fold_right ~f ~init:initial a
+
+  let fold_right = foldRight
+
```
j-m-hoffmann: alias to foldr?

bs/src/tablecloth.ml
Dean177: @j-m-hoffmann did you know about GitHub's [suggestion](https://help.github.com/en/articles/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) feature? Its perfect for one-liner changes like these since I can incorporate your improvements directly in the GitHub UI.

bs/src/tablecloth.ml
Dean177: Taking a 300% performance hit is a pretty big deal so it looks like we have two options
- Revert to Belt / Base 's type signature
- Re-implement foldLeft / foldRight from scratch to have the Elm signatures.

It's actually kinda a big deal, as making it ergonomic means we have to ensure all of the other modules have API's that permit "non-flipped" usage 

e.g. Having `foldLeft` take `(f : 'a -> 'b -> 'b)` means that we should also make sure that the signature for `Set.add` is `'a -> 'a t -> 'a t`.

This means making a commitment to either:
1. t-last like Elm
2. t-first like Belt
3. t-middle aka all of the optional arguments come first, then the `t` then any non-labelled arguments then the labelled arguments aka the approach of Base / Core.

I would be in favour of either 1 or 3, I think being able to consistently use `|>` is important:
- I am not a huge fan of the mental overhead of having to switch between `->` and `|>` when using Belt
- It actually gets a lot of recognition outside of the OCaml ecosystem thanks to the likes of Elixir and some JS proposals.

Reading that back we should probably turn this into a GitHub issue.

Dean177:
I think we should stick with Array being mutable and potentially offer an immutable alternative with a name other than Array.

I figure we need to inter-operate on two fronts: Javascript and native OCaml, both of those have mutable Arrays and having a collection with the same name name but different behaviour would be potentially confusing.

bs/src/tablecloth.ml
pbiggar: Good discussion! Agreed: make an issue and let's get this merged. I'm happy to merge it with the bad-performance order - we can follow up with a grand plan for fixing the performance later.

pbiggar:
OK, happy to go with the mutable version, thanks for the opinion. If you want to go through @j-m-hoffmann's review/suggestions and either fix them or decide not to, then we can get this merged.

bs/src/tablecloth.ml
Dean177: Why the preference of belt over the standard library?

bs/src/tablecloth.ml
Dean177: This is currently using the standard libraries implementation of init

bs/src/tablecloth.ml
pbiggar: Because presumably the Belt ones are more optimized (and similarly, I expect the Jane Street ones are more optimized on the native side of things).

bs/src/tablecloth.ml
j-m-hoffmann: thnx for the tip, will do so next time

bs/src/tablecloth.ml
j-m-hoffmann: you can use initialize which has been implemented further up, this way you get the Base/Belt version no matter which file it is

bs/src/tablecloth.ml
j-m-hoffmann: for me the biggest wart of ocaml is the difference the difference between foldl and foldr.
foldl expects a function of type accumulator -> item -> accumulator
foldr expects an function of type item -> accumulator -> accumulator
Belt did the right thing IMHO and unified both to accumulator -> item -> accumulator
we have the chance to do the same and get the same on native if we use flip once on foldr,
which is worse perfomancewise anyway.

j-m-hoffmann: i like the t-middle approach of Base

native/src/tablecloth.ml
Dean177: Its a slightly different signature (initial rather than init).

native/src/tablecloth.ml
Dean177: But it is listed in the `mli`? Or did you mean its an alias of concatenate?

j-m-hoffmann: https://github.com/darklang/tablecloth/issues/22

native/src/tablecloth.ml
j-m-hoffmann:
```suggestion
let foldr = foldRight
```

native/src/tablecloth.ml
j-m-hoffmann:
```suggestion

  let foldl = foldLeft

```

native/src/tablecloth.ml
Dean177: Oh you mean add the `foldr` alias in addition to `foldRight` and `fold_right`.

This was an intentional deviation from Elms standard library as I felt it was more in keeping with Elms guidelines on API's: https://package.elm-lang.org/help/design-guidelines#use-human-readable-names

In this particular case the Elm docs even go on to explain that `foldr` means "fold from the right".

bs/src/tablecloth.ml
Dean177: Ok, I have created #35

Dean177:
@j-m-hoffmann are you happy your comments have been addressed?

If so I think that means this is ready @pbiggar

pbiggar:
Gonna merge this now, to avoid future conflicts. @j-m-hoffmann, if you wanna comment on this PR we can get the addressed after.

native/src/tablecloth.ml
j-m-hoffmann: im in a rush, gonna read it when i have time. My thought was we have indexedMap/mapi

pbiggar: I think we should remove mapi and all synonyms - have just one way to do it.

PR #32, opened 2019-04-04
Document module Option

jdeisenberg:
Documentation for the Option submodule.

### DIFF
--- a/bs/src/tablecloth.mli
+++ b/bs/src/tablecloth.mli
@@ -1528,30 +1528,13 @@ module Result : sig
 end
 
 (**
-  This module implements the `Option` type, which has a variant for 
-  valid values (`'Some`), and one for invalid values (`'None`).
+  This module provides functions to work with the `option` type,
+  which has a variant for  valid values (`'Some`), and one for
+  invalid values (`'None`).
 *)
 
 module Option : sig
-  (**
-    `type` is the type constructor for an `Option` type. You specify
-    the type of the `Some` variant.
-    
-    ### Example
-    
-    Here is how you would annotate an `Option` variable whose `Some`
-    variant is an integer:
-    
-    ```ocaml
-    let x: (int) Tablecloth.Option.t = Some 3
-    let y: (int) Tablecloth.Option.t = None
-    ```
-    
-    ```reason
-    let x: Tablecloth.Result.t(string, int) = Ok(3);
-    let y: Tablecloth.Result.t(string, int) = Error("bad");
-    ```
-  *)
+
   type 'a t = 'a option
 
   (**
@@ -1565,7 +1548,7 @@ module Option : sig
     
     ### Example
     ```reason
-    let recip = (x: float): Tablecloth.Option.t(float) => {
+    let recip = (x: float): option(float) => {
       if (x == 0.0) {
         None;
       } else {
@@ -1582,7 +1565,7 @@ module Option : sig
     calls, each of which returns an `Option` value.
     
     ```reason
-    let root = (x: float): Tablecloth.Option.t(float) => {
+    let root = (x: float): option(float) => {
       if (x < 0.0) {
         None
       } else {
@@ -1608,7 +1591,7 @@ module Option : sig
     
     ### Example
     ```ocaml
-    let recip (x:float) : float Tablecloth.Option.t = (
+    let recip (x:float) : float option = (
       if (x == 0.0) then
         None
       else
@@ -1624,7 +1607,7 @@ module Option : sig
     calls, each of which returns an `Option` value.
     
     ```ocaml
-    let root (x:float) : float Tablecloth.Option.t = (
+    let root (x:float) : float option = (
       if (x < 0.0) then
         None
       else
@@ -1663,7 +1646,7 @@ module Option : sig
 
   (**
     `orElse(opt_a, opt_b)` returns `opt_b` if it is of the form `Some(x)`;
-    otherwise, it returns `opt_a`.
+    otherwise, it returns `opt_a`. 
     
     (Same as `or_else`.)
     
--- a/bs/src/tablecloth.mli
+++ b/bs/src/tablecloth.mli
@@ -1638,30 +1638,183 @@ module Option : sig
   *)
   val and_then : f:('a -> 'b option) -> 'a option -> 'b option
 
+  (**
+    `or_ opt_a opt_b` (`or_(opt_a, opt_b)` in ReasonML) returns
+    `opt_a` if it is of the form `Some x` (`Some(x) in ReasonML);
+    otherwise, it returns `opt_b`.
+    
+    ### Example
+    
+    ```ocaml
+    or_ (Some 11) (Some 22) = Some 11
+    or_ None (Some 22) = Some 22
+    or_ (Some 11) None = Some 11
+    or_ None None = None
+    ```
+
+    ```reason
+    or_(Some(11), Some(22)) == Some(11);
+    or_(None, Some(22)) == Some(22);
+    or_((Some(11)), None) == Some(11);
+    or_(None, None) == None;
+    ```
+  *)
   val or_ : 'a option -> 'a option -> 'a option
 
+  (**
+    `orElse(opt_a, opt_b)` returns `opt_b` if it is of the form `Some(x)`;
+    otherwise, it returns `opt_a`.
+    
+    (Same as `or_else`.)
+    
+    ### Example
+    
+    ```reason
+    orElse(Some(11), Some(22)) == Some(22);
+    orElse(None, Some(22)) == Some(22);
+    orElse((Some(11)), None) == Some(11);
+    orElse(None, None) == None;
+    ```
+
+  *)
   val orElse : 'a option -> 'a option -> 'a option
 
+  (**
+    `or_else opt_a opt_b` returns `opt_b` if it is of the form `Some x`;
+    otherwise, it returns `opt_a`.
+    
+    (Same as `orElse`.)
+    
+    ### Example
+    
+    ```ocaml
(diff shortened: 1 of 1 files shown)


bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
     -> unit
 end
 
+(**
+  This module implements the `Option` type, which has a variant for 
+  valid values (`'Some`), and one for invalid values (`'None`).
```
pbiggar: I think it's more accurate to say it provides functions to work on the option type. While the `t` exists, I don't think anyone is going to use `Option.t` instead of `option` (nor should they).

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+*)
+
 module Option : sig
+  (**
+    `type` is the type constructor for an `Option` type. You specify
+    the type of the `Some` variant.
```
pbiggar: I dont understand what this says.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    Here is how you would annotate an `Option` variable whose `Some`
+    variant is an integer:
+    
+    ```ocaml
+    let x: (int) Tablecloth.Option.t = Some 3
+    let y: (int) Tablecloth.Option.t = None
+    ```
+    
+    ```reason
+    let x: Tablecloth.Result.t(string, int) = Ok(3);
```
pbiggar: The result is here by accident.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    the type of the `Some` variant.
+    
+    ### Example
+    
+    Here is how you would annotate an `Option` variable whose `Some`
+    variant is an integer:
+    
+    ```ocaml
+    let x: (int) Tablecloth.Option.t = Some 3
+    let y: (int) Tablecloth.Option.t = None
```
pbiggar: I dont think we should encourage people to use `Tablecloth.Option.t`

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    parameter and returns a `Option`, to an `Option` variable `opt`.
+    
+    If `opt` is of the form `Some(x)`, `andThen` returns `f(x)`;
+    otherwise `andThen` returns `None`.
+    
+    (Same as `and_then`.)
+    
+    ### Example
+    ```reason
+    let recip = (x: float): Tablecloth.Option.t(float) => {
```
pbiggar: lets use the built-in option type in these examples.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    parameter and returns an `Option`, to an `Option` variable `opt`.
+    
+    If `opt` is of the form `Some x`, `and_then` returns `f x`;
+    otherwise it returns `None`.
+    
+    (Same as `andThen`.)
+    
+    ### Example
+    ```ocaml
+    let recip (x:float) : float Tablecloth.Option.t = (
```
pbiggar: Let's use option.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    and_then ~f:recip (Some 4.0) = Some 0.25
+    and_then ~f:recip None = None
+    and_then ~f:recip (Some 0.0) = None
+    ```
+    
+    `and_then` is usually used to implement a chain of function
+    calls, each of which returns an `Option` value.
+    
+    ```ocaml
+    let root (x:float) : float Tablecloth.Option.t = (
```
pbiggar: Let's use option.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    
+    root 4.0 |> and_then ~f:recip = Some 0.5
+    root (-2.0) |> and_then ~f:recip = None 
+    root(0.0) |> and_then ~f:recip = None
+    ```
+  *)
   val and_then : f:('a -> 'b option) -> 'a option -> 'b option
 
+  (**
+    `or_ opt_a opt_b` (`or_(opt_a, opt_b)` in ReasonML) returns
```
pbiggar: Might want to point out that this doesn't short-circuit.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    
+    ```ocaml
+    with_default ~default:99 (Some 42) = 42
+    with_default ~default:99 None = 99
+    ```
+  *)
   val with_default : default:'a -> 'a option -> 'a
 
+  (**
+    `foldrValues(opt, xs)` When `opt` is of the form `Some(x)`,
```
pbiggar: We're removing this so no doc required.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    
+    (Same as `foldr_values`.)
+    
+    ### Example
+    
+    ```reason
+    foldrValues(Some(99), [100, 101, 102]) == [99, 100, 101, 102];
+    foldrValues(None, [100, 101, 102]) == [100, 101, 102];
+    ```
+  *)
```
pbiggar: Removing this.

bs/src/tablecloth.mli
```diff
@@ -1527,41 +1527,345 @@ module Result : sig
...
+    
+    (Same as `foldrValues`.)
+    
+    ### Example
+    
+    ```ocaml
+    foldr_values (Some 99)  [100; 101; 102] = [99; 100; 101; 102]
+    foldr_values None  [100; 101; 102] = [100; 101; 102]
+    ```
+  *)
```
pbiggar: Removing this.

bs/src/tablecloth.mli
jdeisenberg: OK; I will get rid of it entirely, since we want people to work with `option` instead.

bs/src/tablecloth.mli
jdeisenberg: Sounds good to me.

bs/src/tablecloth.mli
jdeisenberg: OK will do.

bs/src/tablecloth.mli
jdeisenberg: I am not sure of this wording:

Note that `or_` does not short-circuit; when you call `or_`, both arguments are evaluated before being passed to the function.

How does that sound to you?

bs/src/tablecloth.mli
jdeisenberg: I am not getting rid of this, as another PR will get rid of it for us.

bs/src/tablecloth.mli
jdeisenberg: Not sure how to word this. How does this sound:

Unlike the built in `or` operator, the `or_` function does not short-circuit. When you call the function, both the arguments are evaluated so they can be passed to the function.

jdeisenberg: Oops. For some reason, I didn't see my comment, so I duplicated it. Let me know which version of the wording you prefer.

pbiggar: Either one is fine, thanks!

bs/src/tablecloth.mli
pbiggar: I think I already merged the removal.

bs/src/tablecloth.mli
jdeisenberg: Taken out with preceding change.

bs/src/tablecloth.mli
jdeisenberg: Changed.

bs/src/tablecloth.mli
jdeisenberg: Done.

bs/src/tablecloth.mli
jdeisenberg: Done.

bs/src/tablecloth.mli
jdeisenberg: Done (after next commit/push)

bs/src/tablecloth.mli
jdeisenberg: Removed doc, but not the definition, so that the build can succeed.

msvbg:
This will work with [odoc](https://github.com/ocaml/odoc) if you put the doc comments in the native `.mli` as well, since only the native project builds with dune. Although, I don't think copying and pasting the comments between the two files sounds like a good idea. Why are there two interface files anyway, are they not identical?

pbiggar:
> This will work with odoc if you put the doc comments in the native .mli as well, since only the native project builds with dune. Although, I don't think copying and pasting the comments between the two files sounds like a good idea. Why are there two interface files anyway, are they not identical?

The mlis aren't identical, as we expose different `t` types in StrMap.t etc. We don't want those types to be opaque as you probably want to unify them in some cases, esp while Tablecloth is in it's early stages and hasn't got all the functionality you might hope for just yet.

I don't object to just copying the comments. We did want to have a solution that had a single place to write comments, but we punted on it in favour of just getting some comments written. Now that's we've got a lot more comments it might be a good time to revisit that, though let's not revisit it in this PR, and instead focus on getting it merged. Happy to discuss that in an issue.

pbiggar:
@jdeisenberg I rebased this and fixed the merge conflicts, and I'm about to merge once CI is done. Can you give it a quick check to make sure I didn't break anything? Thanks!

jdeisenberg:
Looks good to me.

PR #41, opened 2019-04-19
Document option

jdeisenberg:
Added documentation for the Option submodule.


jdeisenberg:
Oops. I already documented this. The only difference is an update for a spelling error.

PR #44, opened 2019-04-21
Documentation for Tuple2 and Tuple3 submodules.

jdeisenberg:
Added documentation for these submodules.

### DIFF
--- a/bs/src/tablecloth.mli
+++ b/bs/src/tablecloth.mli
@@ -2028,86 +2028,597 @@ module Int : sig
 end
 
 module Tuple2 : sig
+
+  (**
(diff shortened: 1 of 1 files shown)


Dean177:
These modules are documented in the native/src/tablecloth.mli and it looks like the docstrings you have added are slightly different (apart from adding additional reasonml examples) and in some cases more extensive.

I think we should try to keep the wording consistent between the two `mli` files.

jdeisenberg:
Not sure what to do here. Paul?

pbiggar:
I agree it's worthwhile to address @Dean177's comments.

I'm going to merge as is, and we can address in a followup. My preference is for forward momentum over perfect PRs 😄

PR #47, opened 2019-04-24
Expand the Int and Float modules for 'native'

Dean177:
This expands the Int and Float modules to cover the operations detailed in https://package.elm-lang.org/packages/elm/core/latest/Basics#Int

Some notable differences include:
- All operators also get a named function equivalent
- The `clamp`, `hypotenuse` and `inRange` functions are introduced 
- `round`, `floor`, `ceiling` and `truncate` return a `float` instead of an int with `Float.toInt` serving as the only way to convert to an integer.
- `round` has been expanded to cover (most of) the many possible ways to go about rounding floating point numbers. The documentation for this one might be a little over the top, I wasn't sure about the clearest way to demonstrate the behaviour.
- `logBase` becomes `log ~base:`
- `min`, `max`, `mod`, `sqrt` and `abs` get non-abbreviated names
- `remainderBy` and `modBy` drop the 'By' and get a labelled argument
- `clamp` and `inRange` can throw an exception if the `upper` argument is less than the `lower` argument.

I intend to squash the commits.

### DIFF
--- a/bs/bsconfig.json
+++ b/bs/bsconfig.json
@@ -20,7 +20,8 @@
   "bs-dev-dependencies": [
     "@glennsl/bs-jest"
   ],
-  "bsc-flags": ["-bs-g", "-bs-super-errors", "-warn-error", "@A", "-w", "-4-9-30-40-41-42-102"],
+  "bsc-flags": [
+    "-bs-g", "-bs-super-errors", "-warn-error", "@A", "-w", "-4-9-30-40-41-42-44-102"],
   "warnings": {
     "error" : "+101"
   },
--- a/native/src/tablecloth.ml
+++ b/native/src/tablecloth.ml
@@ -688,20 +688,277 @@ module Char = struct
   let is_printable = isPrintable
 
   let isWhitespace = Base.Char.is_whitespace
-  
+
   let is_whitespace = isWhitespace
 end
 
+module Float = struct
+  type t = float
+
+  let add = (+.)
+
+  let (+) = (+.)
+
+  let subtract = (-.)
+
+  let (-) = (-.)
+
+  let multiply = ( *. )
+
+  let ( * ) = ( *. )
+
+  let divide n ~by = (n /. by)
+
+  let ( / ) = ( /. )
+
+  let power ~base ~exponent = (base ** exponent) 
+
+  let ( ** ) = ( ** )
+
+  let negate = Base.Float.neg
+
+  let (~-) = negate
+
+  let absolute = Base.Float.abs
+
+  let clamp n ~lower ~upper =
+    if upper < lower then
+      raise (
+        Invalid_argument ("~lower:" ^ (Base.Float.to_string lower) ^ " must be less than or equal to ~upper:" ^ (Base.Float.to_string upper))
+      )
+    else if (Base.Float.is_nan lower || Base.Float.is_nan upper || Base.Float.is_nan n) then
+      Base.Float.nan
+    else
+      max lower (min upper n)
+
+  let inRange n ~lower ~upper =
+    if Base.Float.(upper < lower) then
+      raise (
+        Invalid_argument ("~lower:" ^ (Base.Float.to_string lower) ^ " must be less than or equal to ~upper:" ^ (Base.Float.to_string upper))
+      )
+    else
+      n >= lower && n < upper
+
+  let in_range = inRange
+
+  let squareRoot = sqrt
+
+  let square_root = squareRoot
+
+  let log n ~base = Base.Float.(log10(n) / log10(base))
+
+  let zero = 0.0
+
+  let one = 1.0
+
+  let nan = Base.Float.nan
+
+  let infinity = Base.Float.infinity
+
+  let negativeInfinity = Base.Float.neg_infinity
+
+  let negative_infinity = negativeInfinity
+
+  let e = Base.Float.euler
+
+  let pi = Base.Float.pi
+
+  let isNaN = Base.Float.is_nan
+
+  let is_nan = isNaN
+
+  let isInfinite = Base.Float.is_inf
+
+  let is_infinite = isInfinite
+
+  let isFinite n = not (isInfinite n) && not (isNaN n)
+
+  let is_finite = isFinite
+
+  let maximum x y =
+    if isNaN x || isNaN y then
+      nan
+    else if y > x then
+      y
+    else
+      x
+
+  let minimum x y =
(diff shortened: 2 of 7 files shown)


native/src/tablecloth.ml
```diff
@@ -692,16 +692,242 @@ module Char = struct
...
+module Float = struct 
+  type t = float
+
+  let add = (+.)
+
+  let (+) = (+.)
+
+  let subtract = (-.)
+
+  let (-) = (+.)
```
pbiggar: typo

native/src/tablecloth.mli
```diff
@@ -728,34 +728,855 @@ module Char : sig
...
+      1e3        (* = (1 * 10 ** 3) = 1000 *)
+    ]}
+
+    Without opening this module you can use the [.] suffixed operators e.g
+    
+    {[ 1. +. 2. /. 0.25 *. 2. = 17. ]}
+
+    But by opening this module locally you can use the un-suffixed operators
+
+    {[Float.((10.0 - 1.5 / 0.5) ** 3.0) = 2401.0]}
```
pbiggar: This is really cool!

native/src/tablecloth.mli
```diff
@@ -728,34 +728,855 @@ module Char : sig
...
+  (** See {!Float.power} *)
+
+  val negate : t -> t
+  (** Flips the 'sign' of a [float] so that positive floats become negative and negative integers become positive. Zero stays as it is.
+
+    {[Float.negate 8 = (-8)]}
+
+    {[Float.negate (-7) = 7]}
+
+    {[Float.negate 0 = 0]}
```
pbiggar: Sure about this? Isn't the -0 in floats? Can we test this?

native/src/tablecloth.mli
```diff
@@ -728,34 +728,855 @@ module Char : sig
...
+
+  (** {1 Angles} *) 
+
+  val hypotenuse : t -> t -> t
+  (** [hypotenuse x y] returns the length of the hypotenuse of a right-angled triangle with sides of length [x] and [y], or, equivalently, the distance of the point [(x, y)] to [(0, 0)]. 
+
+    {[Float.hypotenuse 3. 4. = 5.]}
+  *)
+
+  val degrees : t -> t
```
pbiggar: Seems an odd name for this function (or perhaps odd functionality)

native/src/tablecloth.mli
```diff
@@ -728,34 +728,855 @@ module Char : sig
...
+  (** Converts an angle in {{: https://en.wikipedia.org/wiki/Degree_(angle) } degrees} to {{: https://en.wikipedia.org/wiki/Radian } radians }.
+
+    {[Float.degrees 180. = 3.141592653589793]}
+  *)
+
+  val radians : t -> t
+  (** Convert radians to {{: https://en.wikipedia.org/wiki/Radian } radians }.
```
pbiggar: I dont understand this docstring. "Convert radians to radians"?

pbiggar:
This is great! Directionally, absolutely continue with it. 

I made a few comments and read over the rest. Nothing I can see that's really out of shape, and the comments are fantastic as well.

native/src/tablecloth.mli
Dean177: That’s correct, but -0.0 = 0.0 evaluates to true.

I intend to turn all of the examples into tests.

native/src/tablecloth.mli
Dean177: The cos and sin functions expect to be called with a value in radians, this (and degrees) are intended (from my understanding of the elm docs) to be used as ‘labels’ e.g ‘cos (radians pi)’.

I didn’t think this was a great part of the elm docs and was considering proposing a radians module.

I will improve the docstring and give some examples of its intended usage.

native/src/tablecloth.mli
Dean177: Unfortunately the example isn’t correct!

pbiggar:
@Dean177 checking in on this?

Dean177:
Sorry to leave it hanging, I have been without laptop for the past week.

bs/bsconfig.json
```diff
@@ -20,7 +20,8 @@
   "bs-dev-dependencies": [
     "@glennsl/bs-jest"
   ],
-  "bsc-flags": ["-bs-g", "-bs-super-errors", "-warn-error", "@A", "-w", "-4-9-30-40-41-42-102"],
+  "bsc-flags": [
+    "-bs-g", "-bs-super-errors", "-warn-error", "@A", "-w", "-4-9-30-40-41-42-44-102"],
```
Dean177: This adds `-44` which disables a warning about shadowing due to a local open.

Dean177:
I think this is ready for some more scrutiny now. 

I did encounter some different behaviour between bucklescript and native in the `Float.toString` function.

The result is there are some tests commented out on the native side, and a note on the mli saying there is a difference, but not going into detail.

I think the best thing to do might be to add a "format" type function which allows precise control over the representation (things like significant digits and trailing zeros etc), but that is probably something to be discussed in an issue rather than bundled into this allready very large PR.

Dean177:
Additionally something i was considering was adding labelled arguments to the `divide` and `power` functions to make them easier to partially apply / understand the argument order e.g.

```ocaml
let divide n ~by = n / by
let halve = divide ~by:2

let power ~base ~exponent = base ** exponent
let square = power ~exponent:2
let powerOfTwo n = power ~base:2 ~exponent:n
```

Thoughts?

pbiggar:
>  labelled arguments to the divide and power

I like it.

pbiggar:
> I did encounter some different behaviour between bucklescript and native in the Float.toString function.

Perhaps comment out for now. It's easy enough for people to use platform-functions while we get that right.

pbiggar:
OK, reviewed it all and it looks great. 

One question I have is about int sizes. Bs uses a 53-bit int and Native uses a 31-bit/63-bit int. Should we unify these? Should we have Int31 and Int53 be modules? Should there still be an Int module? Should we pick a size that works on both?

Dean177:
I think we should take the same approach as Base / Core:
- Have the `Int` module mean "platform specific integer"
- Provide modules such as Int64 / 32 / ... for use cases which need guarantees about integer sizes.

It means you have convenience by default, have the option of using the included `maximumValue` / `minimumValue` if you need to do something a little unusual or going for the IntN modules for complete control.

pbiggar:
OK, this looks good to merge once CI is green.

PR #49, opened 2019-04-30
Fix dune-project location (fixes #45)

### LINKED ISSUE #45: Opam install doesn't work
This appears to be a known issue, but I'm creating an issue in this repository too so it doesn't get lost: https://github.com/ocaml/opam-repository/pull/13670#issuecomment-480545943.

kit-ty-kate:
the dune-project and files are at the wrong place, the `<project>.opam` file and the `dune-project` file should be in the same directory and opam installs the using the `<project>.install` that is on the same level too. The install file taken by opam should be on the toplevel directory so previously, nothing was installed.


pbiggar:
Great, thanks. Will merge once CI is green.

pbiggar:
Quick followon question how should we resolve the opam 0.0.6 package. Do i need to retag this new version as 0.0.6?

PR #53, opened 2019-05-18
Convert docs to OCaml format only.

jdeisenberg:
Following the pattern established for `Float` and `Int`, I have removed the Reason documentation and converted from markdown to `odoc` format.

This is done in anticipation of a time when `odoc` or a similar program will be able to extract sample code and convert it to Reason during the build process.

### DIFF
--- a/bs/src/tablecloth.mli
+++ b/bs/src/tablecloth.mli
@@ -1543,8 +1543,6 @@ module Float : sig
       {[
 let isNotANmber x = Float.(x = nan) in
 isNotANumber nan = false
-
-
 ]}
 
       For detecting [Nan] you should use {!Float.isNaN}
@@ -1572,9 +1570,10 @@ isNotANumber nan = false
   val add : t -> t -> t
   (** Addition for floating point numbers.
 
-    {[Float.add 3.14 3.14 = 6.28]}
-
-    {[Float.(3.14 + 3.14 = 6.28)]}
+    {[
+    Float.add 3.14 3.14 = 6.28
+    Float.(3.14 + 3.14 = 6.28)
+    ]}
 
     Although [int]s and [float]s support many of the same basic operations such as
     addition and subtraction you {b cannot} [add] an [int] and a [float] directly which
@@ -1652,11 +1651,11 @@ isNotANumber nan = false
   val negate : t -> t
   (** Flips the 'sign' of a [float] so that positive floats become negative and negative integers become positive. Zero stays as it is.
 
-    {[Float.negate 8 = (-8)]}
-
-    {[Float.negate (-7) = 7]}
-
-    {[Float.negate 0 = 0]}
+    {[
+    Float.negate 8 = (-8)
+    Float.negate (-7) = 7
+    Float.negate 0 = 0
+    ]}
 
     Alternatively an operator is available:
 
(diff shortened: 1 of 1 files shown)


Dean177:
Looks like it’s on the horizon for odoc:

https://github.com/ocaml/odoc/issues/129

And it’s getting some support to gain this feature

https://www.reason-association.org/projects/

jdeisenberg:
The preceding “consolidate“ commit groups together individual `{[ ... ]}` examples for a function into a single `{[...]}` block.

It also puts the documentation with the BuckleScript snake_case version of functions, and the ReasonML camelCase version becomes `Same as {!Module.see_other}.`

pbiggar:
I know I have this in email somewhere, but for the benefit of others reading this, can you comment about why you think this is the right plan and what the future of this is?

Thanks!

jdeisenberg:
In a conversation with some people involved with odoc, we came to the conclusion that the easiest things to do are to recognize code within examples and extract them to create verifiable tests as part of the document build process. Another step in the process would be to convert the example docs to ReasonML

bs/src/tablecloth.mli
```diff
@@ -7,71 +7,49 @@ Function names that are in snake_case have their documentation written in OCaml
...
 (**
-    The `>>` operator returns a function that is the equivalent of the composition of its function arguments.
-    The main use of `>>` is to avoid writing parentheses.
-
-  `(f >> g) x` (`(f >> g)(x)` in ReasonML) is the equivalent of `f (g x)` (`f(g(x))` in ReasonML)
+    The [>>] operator returns a function that is the equivalent of the composition of its function arguments.
+    The main use of [>>] is to avoid writing parentheses.
 
-  ### Example
+  [(f >> g) x] is the equivalent of [f (g x)]
```
wolfgangGoedel: This should be `[(f >> g) x] is the equivalent of [g (f x)]`

bs/src/tablecloth.mli
```diff
@@ -7,71 +7,49 @@ Function names that are in snake_case have their documentation written in OCaml
...
 
+val ( << ) : ('b -> 'c) -> ('a -> 'b) -> 'a -> 'c
 (**
-  The `<<` operator returns a function that is the equivalent of the reverse composition of its function arguments.
-
-  `(f << g) x` (`(f << g)(x)` in ReasonML) is the equivalent of `g (f x)` (`g(f(x))` in ReasonML)
+  The [<<] operator returns a function that is the equivalent of the reverse composition of its function arguments.
 
-  ### Example
+  [(f << g) x] is the equivalent of [g (f x)]
```
wolfgangGoedel: ...and this the other way: `[(f << g) x] is the equivalent of [f (g x)]`
The docs interchanged signatures for forward and backward composition

pbiggar:
Apologies for the radio silence, i missed this. @jdeisenberg - can you rebase and handle @wolfgangGoedel's comments? Thanks!

jdeisenberg:
I'm not sure how to do a rebase. I have never done one successfully, and always end up removing all the files, cloning again, and starting over.

pbiggar:
I rebased this and fixed the merge conflicts. Can you check over to make sure it's right (the conflicts were in List.concat, which had moved in the file).

Merging this now to avoid future conflicts.

PR #55, opened 2019-05-23
Fix type signatures for flatMap

figitaki:
The interface file doesn't match the types in the implementation file. Small change that fixes the interface file to accurately reflect the type signature of flatMap.


pbiggar:
Great, thanks! Can you do this in the other tablecloth.mli file (in native)?

figitaki:
For sure!

pbiggar:
Thanks!

PR #61, opened 2019-06-22
WIP: Add Float.toString/to_string

kuy:
*(PR code is Work In Progress)*

I wanted to make an implementation that has the exact same behaviour for both platforms, but I finally found it's difficult without author's dicision. For example, a string representation of **Infinity**: In JavaScfipt, `Infinity` is used and `inf` in OCaml. In addition to this, OCaml returns `3.` (with dot) but JavaScript returns just `3`.  [Full output](https://gist.github.com/kuy/0318f923ca36cc42a0992658ef08d95c) at Gist.

- Option 1: Allow different output for each platform
- Option 2: Define an output of Tablecloth (neither JS nor OCaml)
- Option 3: Standardize to JS or OCaml

In [Design of Tablecloth](https://github.com/darklang/tablecloth#tablecloths-solution) section, you say that Tablecloth should has identical API, but it doesn't say Tablecloth should return the exact same output. OCaml users may suprise if `Float.to_string` returns `Infinity` and ReasonML/BackleScript users may suprise if `Float.toString` returns `inf`.

My conclusion: Option 1

Pros: No suprise for both platforms.
Cons: Need to write some code to get the exact same output.

I think Tablecloth need to keep a neutral position. To keep it, **Option 3** is a possible option.

Please let me hear what you think :)

### DIFF
--- a/bs/__tests__/tablecloth_test.ml
+++ b/bs/__tests__/tablecloth_test.ml
@@ -710,6 +710,21 @@ let () =
       test "infinity" (fun () -> expect (toInt infinity) |> toEqual None);
       test "negativeInfinity" (fun () -> expect (toInt negativeInfinity) |> toEqual None);
     );
+
+    describe "toString" (fun () ->
+      test "3." (fun () -> expect (toString 3.) |> toEqual "3");
+      test "-3." (fun () -> expect (toString (-3.)) |> toEqual "-3");
+      test "0." (fun () -> expect (toString 0.) |> toEqual "0");
+      test "-0." (fun () -> expect (toString (-0.)) |> toEqual "0");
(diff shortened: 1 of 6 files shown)


Dean177:
There is also option 4: don’t provide this as part of tablecloth and force users to rely on the platform specific API.

My concern with option one is someone doing Float.toString on the native side, then Float.fromString in JS/Bucklescript and ending up with NaN.

Option four would be a shame though as it is explicitly the goal of TC to smooth over these sorts of things. 

Perhaps # 1 and some documentation explaining the issue.

pbiggar:
My strong preference is Option 2/3 (probably as close to JS as possible). However, is that really possible? If not, I'd opt to go with option 4. How likely do you think it is that we can normalize this?

Dean177:
Float.toString was added in #130 going for Option 1. Its currently under-documented though so the addition of tests and docs would still be very helpful

PR #65, opened 2019-07-22
Add `Option.getExn`

dmnd:
If this was left out of the API for philosophical reasons, feel free to
close the PR, but I think it makes sense. Bikeshedding the name welcome.

See also [this discussion][1] on StackOverflow about the corresponding
API in Elm.

[1]: https://stackoverflow.com/questions/28699800/right-way-to-forcibly-convert-maybe-a-to-a-in-elm-failing-clearly-for-nothings

Test plan:

`./build.sh` passes


pbiggar:
Thanks for the contribution.

I had somewhat intended to be stricter with types and not allow this sort of thing, but since we use it quite a bit, that probably wouldn't make sense :)

PR #66, opened 2019-07-25
Support Result.fromOption

wpcarro:
This is taken from Elm's core Result module:
https://package.elm-lang.org/packages/elm/core/latest/Result

I'm new to OCaml, so if any of the idioms should change or if there are stylistic comments, don't hold back.

Thanks for making this library!

### DIFF
--- a/bs/__tests__/tablecloth_test.ml
+++ b/bs/__tests__/tablecloth_test.ml
@@ -3,6 +3,13 @@ open Jest
 open Expect
 
 let () =
+  describe "Result" (fun () ->
+    describe "fromOption" (fun () ->
+      test "maps None into Error" (fun () -> expect Result.(fromOption ~error:"error message" None) |> toEqual (Belt.Result.Error "error message"));
+      test "maps Some into Ok" (fun () -> expect Result.(fromOption ~error:"error message" (Some 10)) |> toEqual (Belt.Result.Ok 10));
+    );
+  );
+
   describe "Array" (fun () ->
     describe "empty" (fun () -> 
       test "has length zero" (fun () -> expect Array.(empty |> length) |> toEqual 0);
--- a/bs/src/tablecloth.ml
+++ b/bs/src/tablecloth.ml
@@ -441,6 +441,13 @@ module Result = struct
   let map (f : 'ok -> 'value) (r : ('err, 'ok) t) : ('err, 'value) t =
     Belt.Result.map r f
 
+  let fromOption ~(error) ma = 
+    match ma with
+    | None -> fail error
+    | Some right -> succeed right
+
+  let from_option = fromOption
+
   let toOption (r : ('err, 'ok) t) : 'ok option =
     match r with Ok v -> Some v | _ -> None
 
--- a/native/src/tablecloth.ml
+++ b/native/src/tablecloth.ml
@@ -576,6 +576,13 @@ module Result = struct
   let map (f : 'ok -> 'value) (r : ('err, 'ok) t) : ('err, 'value) t =
     Base.Result.map r ~f
 
+  let fromOption ~(error) ma = 
+    match ma with
+    | None -> fail error
+    | Some right -> succeed right
+
+  let from_option = fromOption
+
   let toOption (r : ('err, 'ok) t) : 'ok option =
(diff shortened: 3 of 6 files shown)


bs/src/tablecloth.ml
```diff
@@ -441,6 +441,13 @@ module Result = struct
   let map (f : 'ok -> 'value) (r : ('err, 'ok) t) : ('err, 'value) t =
     Belt.Result.map r f
 
+  let fromOption left ma =
```
wpcarro: I'm guessing we'll want kwargs here instead of positional args. Right?

native/src/tablecloth.ml
```diff
@@ -576,6 +576,13 @@ module Result = struct
   let map (f : 'ok -> 'value) (r : ('err, 'ok) t) : ('err, 'value) t =
     Base.Result.map r ~f
 
+  let fromOption left ma =
```
wpcarro: Same comment about kwargs for here as well I presume.

bs/src/tablecloth.ml
Dean177: Maybe `left` could be renamed (errorMessage?) and optional with a default error message being provided. 

Then the usage would be:
```
Result.fromOption None
```
or
```
Result.fromOption ~errorMessage:"More specific message" None
```

bs/__tests__/tablecloth_test.ml
```diff
@@ -3,6 +3,13 @@ open Jest
 open Expect
 
 let () =
+  describe "Result" (fun () ->
+    describe "fromOption" (fun () ->
+      test "maps None into Error" (fun () -> expect Result.(fromOption "error message" None) |> toEqual (Belt.Result.Error "error message"));
```
Dean177: This should probably be
```suggestion
      test "maps None into Error" (fun () -> expect Result.(fromOption "error message" None) |> toEqual (Result.Error "error message"));
```
That it uses Belt under the covers is an implementation detail.

bs/__tests__/tablecloth_test.ml
```diff
@@ -3,6 +3,13 @@ open Jest
...
+      test "maps Some into Ok" (fun () -> expect Result.(fromOption "error message" (Some 10)) |> toEqual (Belt.Result.Ok 10));
```
Dean177:
```suggestion
      test "maps Some into Ok" (fun () -> expect Result.(fromOption "error message" (Some 10)) |> toEqual (Result.Ok 10));
```

bs/src/tablecloth.ml
wpcarro: I agree `left` is too vague.

The problem I see with `errorMessage` is that `message` implies it's a `string` when it should really be `'a`.

I'll change it to `error` so we can do `Result.fromOption ~error: "error message" None`.

Also since it's `'a` instead of `string`, I don't think it makes sense to provide a default value and narrow the types unnecessarily.

pbiggar:
@Dean177 @wpcarro Looks like everything is handled? Are we good to merge?

wpcarro:
@pbiggar all set for me

bs/__tests__/tablecloth_test.ml
```diff
@@ -3,6 +3,13 @@ open Jest
 open Expect
 
 let () =
+  describe "Result" (fun () ->
+    describe "fromOption" (fun () ->
+      test "maps None into Error" (fun () -> expect Result.(fromOption ~error:"error message" None) |> toEqual (Belt.Result.Error "error message"));
+      test "maps Some into Ok" (fun () -> expect Result.(fromOption ~error:"error message" (Some 10)) |> toEqual (Belt.Result.Ok 10));
```
Dean177: Did you see my suggestions on the previous commit?

pbiggar:
Thanks @wpcarro, and thanks @Dean177 for the review!

bs/__tests__/tablecloth_test.ml
wpcarro: @Dean177 whoops I forgot to update those. What do you want me to do? I can open a PR dropping the `Belt.Result.` namespace.

PR #70, opened 2019-07-26
Support ocamlformat

wpcarro:
- Runs `ocamlformat` across project's source code
- Adds job to .circleci/config.yml to assert that the project is formatted

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,9 +49,25 @@ jobs:
           paths:
             - ~/.opam
 
+  source-code-formatting:
+    docker:
+      - image: ocaml/opam2
+    environment:
+      CI: true
+    working_directory: ~/repo
+    steps:
+      - checkout:
+          path: ~/repo
+      # m4 is a system dependency required by conf-m4 -> ocamlfind -> fmt -> alcotest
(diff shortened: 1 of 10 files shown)


wpcarro:
As I'm pretty new to the OCaml community, I'm unsure if the style that the `.ocamlformat` config is giving us is desirable or not. If there are some options that we'd like to set and others that we'd like to unset, let me know and I can make those changes.

Otherwise, unless I'm mistaken, this should be good to go.

wpcarro:
@pbiggar should be all set to go!

pbiggar:
I don't see the settings in the PR? As a reminder:

> Can you use these settings?

```
profile = sparse
break-string-literals = never
escape-strings = preserve
```

wpcarro:
@pbiggar I believe you added `.ocamlformat` with those exact settings. See [here](https://github.com/darklang/tablecloth/blob/master/.ocamlformat). Unless I'm misunderstanding your ask, we should be ok. No?

wpcarro:
Let me know if there's anything else you'd like me to change before we can submit this. Thanks!

pbiggar:
Thank you!

PR #75, opened 2019-08-02
Add the `Poly` module for `Map` and `Set`. 

Dean177:
This PR adds polymorphic maps and sets as a new submodule to Map and Set.

In addition:
- The functions are documented
- Most functions in Set come with examples 
- IntDict and StringDict become the Int and String submodules of Map and IntSet and StringSet become the Int and String submodules of Set. I did this to be more consistent with both OCaml's and Javascript's terminology (both use 'Map' rather than 'Dict' / 'Dictionary')
- Expands the api provided by the *Dict and *Set modules to be comparable with the ones provided in Elms core.

This is just for the native side. If you approve of the API I will add the bs side and test suite.

### DIFF
--- a/native/src/tablecloth.ml
+++ b/native/src/tablecloth.ml
@@ -576,7 +576,7 @@ module Result = struct
   let map (f : 'ok -> 'value) (r : ('err, 'ok) t) : ('err, 'value) t =
     Base.Result.map r ~f
 
-  let fromOption ~(error) ma = 
+  let fromOption ~(error) ma =
     match ma with
     | None -> fail error
     | Some right -> succeed right
@@ -943,6 +943,8 @@ module Int = struct
 end
 
 module String = struct
+  type t = string
+
   let length = String.length
 
   let toInt (s : string) : (string, int) Result.t =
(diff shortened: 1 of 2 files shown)


pbiggar:
Belt uses non-module based Map/Set/Dict because module based ones have poor performance on JS. At least, that's my understanding. Which I think will be a problem with this interface?

Can you write up a short doc of how to use this, especially with how to use it for new types? I feel that the entire Core/Base library is designed around this interface which makes it more difficult since we are not. But we could be? I don't know.

Dean177:
> Belt uses non-module based Map/Set/Dict because module based ones have poor performance on JS

This structure (sub-modules inside Set / Map) lines up with the way both Belt and Core organise things:
- [Belt.Set.Int](https://bucklescript.github.io/bucklescript/api/Belt.Set.html)
- [Base.Set.Poly](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/Set/index.html#module-Poly)

Did you mean functors?

Belt and Base both provide interfaces that (if your module satisfies them) that allow you to create Set/Map's with custom data types e.g. for Belt 
```ocaml
let customEmptySet = Belt.Set.fromArray [||] (module MyCustomModule)
```
And for Base
```ocaml
let customEmptySet = Base.Set.of_array (module MyCustomModule) [||] 
```

Then both libraries ensure that all of their modules satisfy the relevant interface, and export that interface for your own modules to use too.

> I feel that the entire Core/Base library is designed around this interface.

But that isn't the approach this PR takes, there is no interface to satisfy, it is achieved instead by using the built in `compare` function.

That means there is no ceremony to using it with new data types, you can use the functions in `Map.Poly` or `Set.Poly` with anything, no setup required:
```ocaml
open Tablecloth
type book = { isbn: string }
let books = Set.Poly.fromList [{ isbn="123"}; {isbn="123"}; {isbn="456"}]
(* books : book Set.Poly.t *)

let coordinates = Set.Poly.fromList [(1, 1); (1, 1); (2, 2)]
(* coordinates : (int * int) Set.Poly.t *)
```

pbiggar:
> Did you mean functors?

Ah yes I did. I guess I must have misunderstood what I read then, thanks for clarifying.

For names, can we go back to IntDict, StrDict and make this PolyDict? (and same for Set: IntSet, et).

> But that isn't the approach this PR takes, there is no interface to satisfy, it is achieved instead by using the built in compare function.

That's true for Poly, but I think the value here comes not from Poly (which I was under the impression had weird side effects and was kinda dangerous?), but for making other Dicts easily. The Base library has a lot of modules called like Int63, and you can make an Int63Dict easily cause Int63 has the right shape to satisfy the contract that Base.Map is looking for.

Dean177:
> For names, can we go back to IntDict, StrDict and make this PolyDict? (and same for Set: IntSet, et).
I changed `Dict` to `Map` since both ocaml & javascript (and Base & Belt) use Map. Why no nested modules? 

> That's true for Poly, but I think the value here comes not from Poly (which I was under the impression had weird side effects and was kinda dangerous?), 

There are indeed reasons not to use `compare`.  The `Map.Poly` and `Set.Poly` module could probably do with a detailed explanation of when it's not ok to use. 

> but for making other Dicts easily. The Base library has a lot of modules called like Int63, and you can make an Int63Dict easily cause Int63 has the right shape to satisfy the contract that Base.Map is looking for.

I intended to look into a way to unify the first-class module / functor based approach that allows you to create new specialised modules, but since it is a fair bit more effort as there needs to be an interface for modules to implement, this seemed like a reasonable first step (and I needed some polymorphic maps for my own projects).

Note that with `Tablecloth.Set.Poly` you don't *need* to create a new `Int64Set` module before you want to create a set of `Int64.t`:
```ocaml
(** Using base *)
open Base

module Book = struct
  module T = struct
    type t = { title: string; isbn: string }
    let compare = dotdotdot
    let sexp_of = dotdotdot
  end
  include T
  include Comparator.Make(T)
end

let bookSet = Set.singleton (module Book) { title="Moby Dick", isbn="9781974305032" }

(** Using tablecloth and Set.Poly *)
open Tablecloth

module Book = struct
    type t = { title: string; isbn: string }  
end

let bookSet = Set.Poly.singleton { title="Moby Dick", isbn="9781974305032" }
```

Note that Base provides its own `Set.Poly` since it is convenient.

pbiggar:
I can see the value of adding the Poly dict, though I could imagine us deprecating it later.

I hate the name - I don't find it at all intuitive; can we do better?

I feel pretty strongly about the names as you can probably tell. I don't think Map is a good name for these at all, and I think Dict is a fine name. Other names could be considered but I ran through the obvious set (Map, Hashtable, Hashmap, Table, Associative Array(!)), and I think Dict makes the most sense.

Dean177:
> I hate the name - I don't find it at all intuitive; can we do better?

Do you mean `Map` or `Poly`?

I agree "Dict" is a better name, it tells you what the data structure does better than "Map". But I think the other factors in play make it a more confusing choice here:
- Ocaml & Javascript, the two primary targets, both use Map rather than Dict and the users of those languages will already know what 'Map' is and its what they will go looking for when they need that functionality.
- Base & Belt, the libraries we are unifying already use Map, so for users of those libraries will encounter less friction if they decide to switch to using Tablecloth. 
- Belt already includes a `Dict` module, that is slightly different from its `Map`: https://bucklescript.github.io/bucklescript/api/Belt_Map.Dict.html

Are you happy with the sub-module structure?

Dean177:
@pbiggar can you take a look at my reply, I had some more questions about the issues you raised.

pbiggar:
Sorry for the delay, thanks for poking me!

> > I hate the name - I don't find it at all intuitive; can we do better?

> Do you mean Map or Poly?

I meant poly. 

> I agree "Dict" is a better name, it tells you what the data structure does better than "Map". But I think the other factors in play make it a more confusing choice here:

I feel that Map is a legacy of the mathematical underpinings of FP and I don't find it helpful. Elm's choice of switching to Dict was a good one. I see your arguments and I don't find them particularly compelling, we are after all trying to abstract over the things these libraries don't do well.

One of the reasons to choose Elm's naming is that they put a serious amount of thought into their API design, so by copying their API design we get to piggy back onto that work. We also get to avoid difficult decisions like this :) Elm didn't get everything perfect and where we have a really good reason for stepping away from Elm's decisions we should, but I dont think we have one here.


> Are you happy with the sub-module structure?

No, I think we shouldn't do it. It's weird, and I've always found it weird wherever it comes up in OCaml.

PR #77, opened 2019-08-02
Add Array.includes

Dean177:


pbiggar:
I think this is called contains in List and String - can you check and make sure this all lines up?

Dean177:
Ah:
Set, List -> member
String -> contains

PR #79, opened 2019-08-02
Add the Fun module

Dean177:
This PR Adds a `Fun` module providing some useful functions (and providing a home for a couple of existing ones).

This is initially just for the native side, but if the API looks good I will add tests and the bs side of things.

### DIFF
--- a/native/src/tablecloth.mli
+++ b/native/src/tablecloth.mli
@@ -1,5 +1,5 @@
 module Fun : sig
-  (** Functions for working with functions.  *)
+  (** Functions for working with functions. *)
 
   external identity : 'a -> 'a = "%identity"
   (** Given a value, returns exactly the same value. This may seem pointless at first glance but it can often be useful when an api offers you more control than you actually need.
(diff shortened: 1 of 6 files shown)


native/src/tablecloth.mli
```diff
@@ -1,10 +1,147 @@
...
+  val composeRight : ('a -> 'b) -> ('b -> 'c) -> 'a -> 'c
+  (** Function composition, passing results along in the suggested direction. For
+example, the following code checks if the square root of a number is odd:
+
+    {[Int.squareRoot >> Int.isEven >> not]} *)
+
+  val ( >> ) : ('a -> 'b) -> ('b -> 'c) -> 'a -> 'c
+  (** See {!Fun.composeRight} *)
+
+  val tap : 'a -> f:('a -> unit) -> 'a
```
pbiggar: What does tap mean?

Dean177: It means "tap into"

Its based on:
- https://apidock.com/ruby/Object/tap
- https://lodash.com/docs/4.17.15#tap

Dean177:
@pbiggar this is now complete for bs & native if you are happy with the naming / docs / API

pbiggar:
Nice work, thank you!

PR #84, opened 2019-08-12
Fix documentation build error

joesiewert:
Getting the following error when running `make documentation` for the first time since the `docs` directory has not been created before:

```
Compiling the documentation ...
opam exec -- dune build @doc
rm ./docs
rm: ./docs: No such file or directory
make: *** [documentation] Error 1
```

Switching to `rm -rf ./docs` should clear it up.


pbiggar:
Thanks!

PR #88, opened 2019-09-22
Remove deprecation warnings for string functions

msvbg:
I am getting warnings like this with bsb 6.0.1:
```ocaml
  1218 ┆ let capitalize (s : string) : string = String.capitalize s
  1219 ┆
  1220 ┆ let isCapitalized (s : string) : bool = s = String.capitalize s
  1221 ┆
  1222 ┆ let is_capitalized = isCapitalized

  deprecated: String.capitalize
Use String.capitalize_ascii instead.
```
This removes the warnings.

### DIFF
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -1,0 +1,1 @@
+bs/src/compat.ml
--- a/bs/bsconfig.json
+++ b/bs/bsconfig.json
@@ -4,7 +4,10 @@
   "sources": [
     {
       "dir" : "src",
-      "subdirs" : false
+      "subdirs" : false,
+      "public": [
+        "Tablecloth"
+      ]
     },
     {
       "dir": "__tests__",
--- a/bs/src/compat.ml
+++ b/bs/src/compat.ml
@@ -1,0 +1,30 @@
+(* This file is used to target different OCaml versions using the BuckleScript
+   preprocessor. This is a separate file since ocamlformat does not accept
+   the BuckleScript preprocessor. *)
+
+module String = struct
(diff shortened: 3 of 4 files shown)


msvbg:
Hm, it appears that the `_ascii` functions don't exist in BuckleScript 5.

msvbg:
It is possible to [target multiple versions of BuckleScript](https://bucklescript.github.io/bucklescript/Manual.html#_examples_3), if you are open to adding a little bit of noise to the code. In this case the CI should probably run once for each version as well.

pbiggar:
I'd be happy to look at conditional support of both versions.

Do we need something that works in the native tests as well (or is there a different thing we want there if you know)?

pbiggar:
Looks good. Can you fix the formatting issues? (We check the files will pass ocamlformat)

msvbg:
@pbiggar Tablecloth Native defers to Base so that should be fine as-is.

I'm not sure I _can_ fix the formatting issues... `#if ...` is not standard OCaml, so ocamlformat does not like it. Looking into it.

pbiggar:
If there isn't a solution, we could isolate the conditional stuff to a single "Compat.ml" file and then exclude that file from the formatter.

msvbg:
Done!

msvbg:
Hold on, may need one more change here.

msvbg:
I think this is correct, based on https://dev.to/yawaramin/bucklescript-best-practice-public-and-private-modules-5ald.

pbiggar:
Nicely done, thanks! As a followup, I'd love to see CI test this on both old and new versions of BS.

PR #90, opened 2019-09-29
Add List.repeat function

msvbg:
As seen in [Elm](https://package.elm-lang.org/packages/elm/core/1.0.2/List#repeat).


pbiggar:
Lovely, thanks!

PR #115, opened 2020-05-06
Make Result.map consistent with other maps

### LINKED ISSUE #114: Result.map inconsistent with Option.map
Option.map uses a labelled parameter for the function and therefore can be called either subject first or subject last (meaning it is interchangeable with `->` and `|>`). However, `Result.map` does not use a labelled parameter for the function meaning it can only be called subject-last, meaning it only works with `|>`. Is there a reason for this inconsistency? Would you accept a PR updating Result to work like Option?

joefiorini:
Adds the `f:` label to `Result.map` in order to make it work either subject-first or subject-last. Fixes #114.

### DIFF
--- a/bs/__tests__/tablecloth_test.ml
+++ b/bs/__tests__/tablecloth_test.ml
@@ -10,7 +10,12 @@ let () =
               |> toEqual (Belt.Result.Error "error message")) ;
           test "maps Some into Ok" (fun () ->
               expect Result.(fromOption ~error:"error message" (Some 10))
-              |> toEqual (Belt.Result.Ok 10)))) ;
+              |> toEqual (Belt.Result.Ok 10))) ;
+      describe "map" (fun () ->
+          test "maps value on success" (fun () ->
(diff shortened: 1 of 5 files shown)


joefiorini:
@pbiggar Any idea why the "source-code-formatting" step of the "buld" workflow failed?

pbiggar:
> @pbiggar Any idea why the "source-code-formatting" step of the "buld" workflow failed?

I don't know the problem but there's a `make format` in the `Makefile` that should fix it.

pbiggar:
> I don't know the problem but there's a make format in the Makefile that should fix it.

Sorry, I hadn't looked at the error message. Weird. Are you able to debug using CircleCI's ssh feature? If not, I can try and look.

pbiggar:
I pushed a possible fix. Also working on a branch that makes it much clearer what wouldn't format properly in CI.

pbiggar:
OK, got the formating fixed, thanks again!

Issue #71, opened 2019-07-26
Dedup test cases?


wpcarro:
This is sort of a wishlist / half-baked idea. Is there a clever way for us to share the test cases between `bs` and `native`?

This may be over-engineered, but perhaps an S-expression DSL that could compile into test cases would be simple enough for us to seriously DRY up the testing code and ensure that our assertions are aligned between `bs` and `native`.

Or maybe there's a better way to design this that wouldn't involve writing a DSL compiler if it were just native OCaml.

Here's a sketch of a datastructure that I think could help dedup a lot of the assertions:

```
(List
  (init
    ("returns None for an empty list" [] None)
    ("returns an empty list for a single item list" ['a'] (Some []))
    ("returns all but the last item for a list with two or more items" ['a';'b';'c';'d'] ['a';'b';'c'])))
```

Like I said, this is half-baked, and I think an OCaml solution would be preferable. Just wanted to raise the issue before forgetting.

Dean177:
I think you are suggesting a tablecloth-test library.

The differences between the bs-jest and [rely](https://www.npmjs.com/package/@reason-native/rely) are already pretty slim. 

Perhaps just porting from alcotest to rely?

wpcarro:
@Dean177 I'm not too worried about the inconvenience of writing tests twice. It's more than duplicating the tests allows for the possibility of different levels of coverage between `bs` and `native`.

So unless there's a solution for fully de-duplicating the tests, I don't have much of an opinion.

I'll look into an OCaml solution and PR it just to illustrate what I have in mind.

dmnd:
If the goal of tablecloth is to be able to write code that works in both native and js environments, it should be possible to write tests in terms of tablecloth and have those tests work in both environments too.

Dean177:
We now have `AlcoJest` which means we can share the tests between implementations

PR #124, opened 2020-05-18
Prep for merge with Standard

### LINKED ISSUE #101: bs-platform 7.1.0 not compatible
```
 
  Warning number 3 (configured as error) 
  node_modules/tablecloth-bucklescript/src/tablecloth.ml 806:47-63
  
  804 ┆ let capitalize (s : string) : string = String.capitalize s
  805 ┆ 
  806 ┆ let isCapitalized (s : string) : bool = s = String.capitalize s
  807 ┆ 
  808 ┆ let is_capitalized = isCapitalized
  
  deprecated: String.capitalize
Use String.capitalize_ascii instead.
  
  We've found a bug for you!
  node_modules/tablecloth-bucklescript/src/tablecloth.ml
  
  The implementation node_modules/tablecloth-bucklescript/src/tablecloth.ml
       does not match the interface src/tablecloth.cmi:
       ...
       In module Array:
       Values do not match:
         val empty : '_a array
       is not included in
         val empty : 'a array
       File "node_modules/tablecloth-bucklescript/src/tablecloth.mli", line 77, characters 2-22:
         Expected declaration
       File "node_modules/tablecloth-bucklescript/src/tablecloth.ml", line 12, characters 6-11:
         Actual declaration
```

Dean177:
The primary thing this pull request does is move the dune-project file to the repository root.

This allows the native implementations `dune` files to reference files from the bucklescript implementation. 

This allows all commands (npm build / dune test) to be run from the repository root, which is slightly more convenient when working on the repository.

Additionally this 
- adds the Bool module & its test cases
- the `AlcoJest` module which allows test to be written in a platform independent way (and so shared between both implementations
- renames the `bs` folder  to `bucklescript`
- removes `empty` from `Array`. Im not sure what this would be used for and it will resolve #101

The rough plan is to do a PR per module, resolving any inconsistencies between the API's as they come up.


bsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "tablecloth",
+  "version": "0.0.6",
```
pbiggar, line 3: this should say 0.0.7.

bsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+  "sources": [
+    {
+      "dir": "bucklescript/src"
```
pbiggar, line 6: Ddi you deliberately drop these fields?
```
      "subdirs" : false,	
      "public": [	
        "Tablecloth"	
      ]
```

package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "tablecloth-bucklescript",
+  "description": "A portable standard library enhancement for Reason and OCaml.",
+  "version": "0.0.6",
```
pbiggar, line 4: This version should be 0.0.7

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tablecloth-bucklescript",
-  "version": "0.0.7",
+  "version": "0.0.6",
```
pbiggar, line 3: Wrong version

bsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "tablecloth",
```
pbiggar, line 2: Why the name change here?

.circleci/config.yml
```diff
@@ -14,17 +14,16 @@ jobs:
...
-            - v1-dependencies-{{ .Branch }}-
-            - v1-dependencies-
-      - run: make deps-bs
-      - run: make build-bs
-      - run: make test-bs
-      - run: make doc-bs
+            - v1-bucklescript-dependencies-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - v1-bucklescript-dependencies-{{ .Branch }}-
+            - v1-bucklescript-dependencies-
+      - run: npm install
```
pbiggar, line 20: I would prefer to keep using the `Makefile` here.

Makefile
```diff
@@ -5,7 +5,7 @@ build-native:
 
 build-bs:
 	@printf "\n\e[31mBuilding bs ...\e[0m\n"
-	cd bs && npm run build
+	cd bucklescript && npm run build
```
pbiggar, line 8: If we flattened the directory structure, we shouldn't need `cd bucklescript` I would think?

Dean177: We don't, I didn't notice since both npm and dune don't mind being run from a sub-directory

bsconfig.json
Dean177: This, and the other strange changes were the result of some careless rebasing, sorry.

Dean177: That said, how come you went for `tablecloth-bucklescript` and `tablecloth-native`?

I was going to bring this up aaaages ago but kinda forgot about it (then the whole Standard thing happened) and actually grabbed the `tablecloth` npm package just in case.

Dean177:
Ah, this closes #71 too

PR #128, opened 2020-05-20
Port the Int module

Dean177:
and share the Interface file and test suites between implementations


Dean177:
There is one tiny change, the floating point division operator is changed from `//` to `/.` since to use the original in reason syntax you would need to do `3 \/\/ 4` as it conflicts with the comment syntax

PR #130, opened 2020-05-20
Port the Char and Float modules & test suites

Dean177:


Dean177:
Yep, just additions and moving to share the interface files

PR #132, opened 2020-05-20
Port the Fun module.  …

Dean177:
Move negate out of bool and curry / uncurry out of Tuple / Tuple3 and into Fun


pbiggar:
I think there's no behaviour change for the existing functions, is that right?

PR #134, opened 2020-05-21
Port the Option and Result modules

Dean177:
Provide the snake case versions of the functions in Option and Result.

Notable changes:
- `Option.getWithDefault` becomes `get ~default` and `|?` 
- `Option.values` gets moved to `List` (and array gains its counterpart)
- `Option.toOption` has been dropped. I have no idea what it was for, it doesn't come from [Elm's  `Maybe`](https://package.elm-lang.org/packages/elm/core/latest/Maybe)
- `Result.t` now matches the ordering of the built in `result` (ok first)
- `Result.succeed` becomes `ok`
- `Result.fail` becomes `error`
- `Result.getWithDefault` becomes `get ~default` and `|?` 
- `Result.combine` became `values`
- `Result.pp` had its arguments swapped to match the type


pbiggar:
> Notable changes:
 
> * `Option.getWithDefault` becomes `get ~default` and `|?`

This doesn't seem right to me. I've found `withDefault` to be pretty clear, but `get` doesn't help me understand what this does very well - I can figure it out via process of elimination ("what else could it do") but it doesn't convey the actual meaning we're looking for. Base uses `Option.value` which honestly I don't find particularly clear either. I'd prefer to stick with `withDefault` here I think.

> * `Option.values` gets moved to `List` (and array gains its counterpart)

I assume this was because this way we can choose the return type by choosing the module to call values from? I don't find this particularly clear though.

> * `Option.toOption` has been dropped. I have no idea what it was for, it doesn't come from [Elm's  `Maybe`](https://package.elm-lang.org/packages/elm/core/latest/Maybe)

No idea!

> * `Result.t` now matches the ordering of the built in `result` (ok first)

This is the OCaml builtin? Sounds good. I've found it very confusing to reason about which is which.

> * `Result.succeed` becomes `ok`
> * `Result.fail` becomes `error`

Nice.

> * `Result.getWithDefault` becomes `get ~default` and `|?`

Same comment as for Option - I think this is better as `Result.withDefault`.

> * `Result.combine` became `values`

I think this is confusing, partially because it's in contrast with the other use(s) of values - I imagine this returns a list of just the `Ok` values, not a Result with a combined set of `Ok` values. I find `combine` to be a good name for this.

> * `Result.pp` had its arguments swapped to match the type

Great!


Other changes:
- going back to our previous discussion of `forEach` - this really makes no sense to me. I get that we want a single word for "doing an action to the thing inside the box" but `forEach` only makes sense on lists.
- `Option.equal`: not against it, but I don't really get it.
- `Option.(>>=)` and `Option.(>>|)`: I don't use these much - can you confirm they work the same as the OCaml builtins?
- `Result.attempt` - omg this is sooooo good! I never thought about this but it's an amazing idea!
- `Result.transpose` - from the name, I assumed this would convert a `(a,b) Result.t` into a `(b, a) Result.` (which wouldn't actually be possible). I guess I'm saying this confuses me. What's it used for?
- `Result.forEach` - same comment as above

Dean177:
> > Notable changes:
> 
> > * `Option.getWithDefault` becomes `get ~default` and `|?`
> 
> This doesn't seem right to me. I've found `withDefault` to be pretty clear, but `get` doesn't help me understand what this does very well - I can figure it out via process of elimination ("what else could it do") but it doesn't convey the actual meaning we're looking for. Base uses `Option.value` which honestly I don't find particularly clear either. I'd prefer to stick with `withDefault` here I think.



> > * `Option.values` gets moved to `List` (and array gains its counterpart)
> 
> I assume this was because this way we can choose the return type by choosing the module to call values from? I don't find this particularly clear though.

Thats the idea, its saves the consumer from having to do something like `Option.values (Array.toList array) |> Array.fromList`

> > * `Result.combine` became `values`
> 
> I think this is confusing, partially because it's in contrast with the other use(s) of values - I imagine this returns a list of just the `Ok` values, not a Result with a combined set of `Ok` values. I find `combine` to be a good name for this.

I agree

> Other changes:
> 
> * going back to our previous discussion of `forEach` - this really makes no sense to me. I get that we want a single word for "doing an action to the thing inside the box" but `forEach` only makes sense on lists.
> * `Result.forEach` - same comment as above

Ok, I think this is falls into the 'doesn't have to have the same name for the same type signature' category.

I don't love `exec` though. `run` is the first thing that occurred to me
> * `Option.(>>=)` and `Option.(>>|)`: I don't use these much - can you confirm they work the same as the OCaml builtins?
They do, I was debating wether or not to include them:
- I find they are a little harder to understand than their named counterparts
- Although they can make code more concise, I think binding operators do the same job in a *much* better way, but they are native-only for now

> * `Result.transpose` - from the name, I assumed this would convert a `(a,b) Result.t` into a `(b, a) Result.` (which wouldn't actually be possible). I guess I'm saying this confuses me. What's it used for?

I don't think it's particularly useful, lets drop it.

> * `Option.equal`: not against it, but I don't really get it.
The `equal` and `compare` methods that have been added to all of the modules do two things:
- Let you easily build compare functions for custom types, which in turn means they are easy to use with maps / sets e.g.
```ocaml
module Maybe3d = struct
  type t = (int * int * int option)
  let compare = Tuple3.compare Int.compare Int.compare (Option.compare Int.compare)
end
```

Dean177:
> > * `Option.getWithDefault` becomes `get ~default` and `|?`
> 
> This doesn't seem right to me. I've found `withDefault` to be pretty clear, but `get` doesn't help me understand what this does very well - I can figure it out via process of elimination ("what else could it do") but it doesn't convey the actual meaning we're looking for. Base uses `Option.value` which honestly I don't find particularly clear either. I'd prefer to stick with `withDefault` here I think.

The named argument (which I think is really useful) makes the name very redundant though `withDefault default` means you end up with `default` twice (and `with_` would be a confusing name).

Surveying some other languages
- Scala & Java call this `get` / `getOrElse`
- Rust calls it `unwrap_or_default`
- Haskell doesn't go there?

I'm not particularly tied to `get` though. I don't think `value` would be the worst choice since its already in Ocaml's Belt& Base. 

Maybe `unwrap`?

pbiggar:
> Ok, I think this is falls into the 'doesn't have to have the same name for the same type signature' category.
> 
> * Belt / Scala calls this `forEach`
> * Elm doesn't have it
> * Base / Ocaml and Rust call it `iter`  🙅‍♂️
>
> I don't love `exec` though. `run` is the first thing that occurred to me, but still doesn't fit quite right.

Yeah, this makes sense to me. I also don't love `exec` or `run`. `do`? 

 
> The `equal` and `compare` methods that have been added to all of the modules do two things:

Thanks, looks great.

pbiggar:
> > > * `Option.getWithDefault` becomes `get ~default` and `|?`
> > 
> > 
> > This doesn't seem right to me. I've found `withDefault` to be pretty clear, but `get` doesn't help me understand what this does very well - I can figure it out via process of elimination ("what else could it do") but it doesn't convey the actual meaning we're looking for. Base uses `Option.value` which honestly I don't find particularly clear either. I'd prefer to stick with `withDefault` here I think.
> 
> The named argument (which I think is really useful) makes the name very redundant though `withDefault default` means you end up with `default` twice (and `with_` would be a confusing name).
> 
> Surveying some other languages
> 
> * Scala & Java call this `get` / `getOrElse`
> * Rust calls it `unwrap_or_default`
> * Haskell doesn't go there?
> 
> I'm not particularly tied to `get` though. I don't think `value` would be the worst choice since its already in Ocaml's Belt& Base.
> 
> Maybe `unwrap`?

I have found `withDefault` to be very clear when using it, despite the annoying duplication. No objections to `unwrap` though.

Dean177:
> > Ok, I think this is falls into the 'doesn't have to have the same name for the same type signature' category.
> > 
> > * Belt / Scala calls this `forEach`
> > * Elm doesn't have it
> > * Base / Ocaml and Rust call it `iter`  🙅‍♂️
> > 
> > I don't love `exec` though. `run` is the first thing that occurred to me, but still doesn't fit quite right.
> 
> Yeah, this makes sense to me. I also don't love `exec` or `run`. `do`?

`do` is a keyword unfortunately

pbiggar:
I've found that `Option.get` has grown on me in the last few days. Still not totally sold on any solution, but I think I'm no longer against it.

Dean177:
> I've found that `Option.get` has grown on me in the last few days. Still not totally sold on any solution, but I think I'm no longer against it.

Ok, if that's the case we just need to figure out a replacement for `forEach` (I don't mind either `get` or `unwrap` so we can go with whichever you prefer) and then this can be merged.

pbiggar:
I asked internally and got one vote for `get`, one for `unwrap`, one for `value` (which was not in the poll). So I think either is fine, you choose.

Dean177:
Of course

Dean177:
How about `Option.tap` instead of `Option.forEach`? This would fit with `Function.tap` (run some side effect in a pipeline) and its gets used for this purpose in JS-land for promises

pbiggar:
> How about `Option.tap` instead of `Option.forEach`? This would fit with `Fun.tap` (run some side effect in a pipeline) and its gets some usage for this purpose in JS-land for [promises](http://bluebirdjs.com/docs/api/tap.html)

Not against it, but have a slight preference for other options. I'd say my order of preference (best->worse) is exec, run, tap.

Dean177:
I put this out as a poll to the people on [reasonml.chat ](https://reasonml.chat/t/tablecloth-standard-merge-new-api-naming-poll/2352)

pbiggar:
> I put this out as a poll to the people on [reasonml.chat ](https://reasonml.chat/t/tablecloth-standard-merge-new-api-naming-poll/2352), I thought it might help is reach a decision

Funny, `tap` was doing really well at the start when I saw it first. Now `foreach` is trending, which is the only option I think is actively bad. I'd suggest we go with `tap` then.

PR #135, opened 2020-05-21
Port the Tuple & Tuple3 modules

Dean177:
The only non-additive changes are `create` became `make`, and `Tuple2` became `Tuple`.


pbiggar:
> The only non-additive changes are `create` became `make`, and `Tuple2` became `Tuple`.

Why change it to Tuple?

I think we should keep it as `Tuple2`. Tuples are any length and `Tuple` does not imply length 2. If you're looking to remove the name, maybe `Pair` is what we want, but I like the symmetry of Tuple2 and Tuple3.

Dean177:
Fair, I will change the module name back to Tuple2

PR #136, opened 2020-05-21
Port the Array & Container modules

Dean177:
Notable changes
- The removal of `empty`, Bucklescript made a breaking change when they made this weakly typed, and since arrays have a fixed length its not very useful anyway
- `initialize` no longer has a `length` label (but retains `f`)
- `floatSum` is removed and `sum` now accepts a first class module
- `concatenate` became `flatten`
- `foldLeft` became `fold`
- `fold` and `foldRight` had the signature changed for `f` its now `f:('b -> 'a -> 'b)` in both cases
- `reverseInPlace` has been removed, `reverse` now returns `unit`. Often doing `|> Fun.tap(~f=Array.reverse)` is what you want anyway


pbiggar:
Some thoughts:

- `Array.get` can throw an exception so perhaps calls this `Array.getUnsafe`? Means that `getAt` could become `get`. Though there's something particular about the `get` function that corresponds to the `.[]` syntax.
- `.?()` - I dont think we need this as there is the `.[]` syntax, right?
- `Array.set` - similar thoughts as for `get`
` Array.includes` - can we call this `Array.contains` (which is actually what the comment says)
- `sum` - very cool!
- > fold and foldRight had the signature changed for f its now f:('b -> 'a -> 'b) in both cases
- I have previously found that having different signatures makes it clearer which one we're trying to use.

Dean177:
> * `Array.get` can throw an exception so perhaps calls this `Array.getUnsafe`? Means that `getAt` could become `get`.  Though there's something particular about the `get` function that corresponds to the `.[]` syntax.

There is, defining `Array.get` changes the behaviour of the `.()` index operator:
``` ocaml
let numbers = [|1; 2; 3|]
let first = Array.get numbers 0 (* val first : int = 1 *)
let second = numbers.(1) (* val second : int = 2 *)

module Array = struct
  let get array index =
    if index < 0 || index > Array.length array 
    then None
    else Some numbers.(index)    
end

let first = Array.get numbers 0 (* val first : int option = Some 1 *)
let second = numbers.(1) (* val second : int option = Some 2 *)
```

> * `.?()` - I don't think we need this as there is the `.[]` syntax, right?
`.[]` is for strings, `.()` is for arrays. The idea is `.()` preserves it's default behaviour and `.?()` is a safer alternative. 

>   ` Array.includes` - can we call this `Array.contains` (which is actually what the comment says)

`contains` is new terminology though 
- Elm doesn't provide this function on array's or lists 
- Js uses includes
- Ocaml calls it `mem` 

> * > fold and foldRight had the signature changed for f its now f:('b -> 'a -> 'b) in both cases
>   * I have previously found that having different signatures makes it clearer which one we're trying to use.
I think there are some downsides to re-ordering for foldRight though:
- It means we will have worse performance (or have to re-implenent foldRight rather than use Base / Belt)
- It means that users need to create an extra function when they what to do things like push to a collection `foldRight ~f:List.cons` vs `foldRight ~f:(fun element accumulator -> List.cons accumulator element)`


> * An example of how to use Container (within Container.ml) would be helpful
For sure

pbiggar:
> > * `.?()` - I don't think we need this as there is the `.[]` syntax, right?
> >   `.[]` is for strings, `.()` is for arrays. The idea is `.()` preserves it's default behaviour and `.?()` is a safer alternative.

Nice!

> > ` Array.includes` - can we call this `Array.contains` (which is actually what the comment says)
> 
> `contains` is new terminology though

It's intended to mirror `List.contains`.
 

> > * > fold and foldRight had the signature changed for f its now f:('b -> 'a -> 'b) in both cases
> >   
> >   
> >   
> >   * I have previously found that having different signatures makes it clearer which one we're trying to use.
> >     I think there are some downsides to re-ordering for foldRight though:
> 
> * It means we will have worse performance (or have to re-implenent foldRight rather than use Base / Belt)
> * It means that users need to create an extra function when they what to do things like push to a collection `foldRight ~f:List.cons` vs `foldRight ~f:(fun element accumulator -> List.cons accumulator element)`

Ignore me then, this is fine.

Dean177:
> It's intended to mirror List.contains.

List currently uses `member` I changed that too!

Dean177:
An email conversation transcript so we have the full history here: 

> I changed List.member to List.includes and added the corresponding function to the Array module too. 

> Elm also calls this member for lists, which I think is not the best name (its terminology typically used in a mathsy context for sets) so I went with the name Javascript uses.

> You proposed ‘contains’ as an alternative, could you expand on why?

> Agree that `member` sucks: I just noticed we've changed Dark's version from `List::contains` to `List::member` and I hate it.

> I think this is one of those subjective things where people's different experiences and vernacular they use come into effect, so I'm not sure what the right thing to do here is. I feel that `List.contains` "feels right", if that makes sense? And `List.includes` doesn't. I suppose in terms of justification lists aren't really about membership (implies by includes: "oh did we include 5") but are just things that are there, and so it makes sense to me that the question is more about is it a happenstance thing: does it contain it. But that's backfilling - I don't really know why `includes` feels so wrong to me.

Dean177:
@pbiggar I can see where you are coming from, lists and are arrays are 'containers' so it makes sense to ask if they contain something. 

They are pretty much interchangeable  but I think that since `includes` is an established name and it fits the other criteria we look for (makes sense in a sentence, is a whole word etc) we should prefer it over `contains`.  

I think if you have still find `includes` too objectionable we could put it to a poll as we have the other APIs, but otherwise I will merge this PR tomorrow as there are others which depend upon it.

pbiggar:
Go for it, thanks. I'll learn to adapt.

Issue #117, opened 2020-05-08
List.insertAt deletes list is inserted longer than the current list


pbiggar:

Dean177:
This was resolved in #137

PR #137, opened 2020-05-21
Port the List module

Dean177:
You should take a look at this after #136

Notable changes

- `concat` becomes `flatten`
- `floatSum` has been removed, `sum` now accepts a module
- `indexedMap` becomes `mapWithIndex` & `mapi` has been removed
- `getBy` is removed leaving just `find`
- `elemIndex` has been replaced by `findIndex`
- `member` becomes `includes`
- `uniqueBy` has been removed (it will be re-introduced as `unique` when I do the PR for `Set`
- `minimumBy` becomes `minimum ~compare`, `minimum` has been removed  (likewise `maximumBy`)
- `sortWith` has been removed, `sortBy` becomes `sort ~compare`
- `span` becomes `splitWhen`
- `iter` becomes `forEach`


pbiggar:
Looks great. THoughts:

- we found that `updateAt` was broken if you tried to add something outside the list (I think the result was the empty list). Have you made any changes or is this still like that?

- `includes` - I think I said this in the other PR but I feel strongly that `includes` isn't a good name here. `List.contains` would be my choice.

pbiggar:
> bucklescript/src/TableclothListExtn.todo.ml 

Stray file?

Dean177:
> Looks great. THoughts:
> 
> * we found that `updateAt` was broken if you tried to add something outside the list (I think the result was the empty list). Have you made any changes or is this still like that?
Yep, I noticed this and fixed them implementation, there is a test case for it too ("updateAt after end of list")

> * `includes` - I think I said this in the other PR but I feel strongly that `includes` isn't a good name here. `List.contains` would be my choice.
Yep, replied to this in the other PR

Dean177:
Fixes #117

PR #144, opened 2020-05-28
Add String.split failing test

tchak:
I run in to this issue in Dark I think.

`String.split` in some edge cases produces different result in `native` and in `bs` implementations.

`make test-bs` -> OK
`make test-native` -> KO

```
ASSERT split - start
[failure] Error split - start: expecting
[; b], got
[b].
Raised at file "stdlib.ml", line 29, characters 17-33
```

In Dark

Screenshot 2020-05-27 at 22 57 13
Screenshot 2020-05-27 at 22 57 25

I found the relevant bit in ocaml docs:
https://caml.inria.fr/pub/docs/manual-ocaml/libref/Str.html#VALsplit

There is `Str.split` and `Str.split_delim`. The second one behaves in the similar manner to JavaScript split used by `bs`. I am not 100% sure which behaviour you want but it should be the same in both implementation. If you want me to change `bs` implementation instead let me know.

### DIFF
--- a/bucklescript/test/StringTest.re
+++ b/bucklescript/test/StringTest.re
@@ -98,6 +98,20 @@ let suite =
     test("reverse", () => {
       expect(String.reverse("stressed")) |> toEqual(Eq.string, "desserts")
     });
+    describe("split", () => {
+      test("middle", () => {
+        expect(String.split("abc", ~on="b"))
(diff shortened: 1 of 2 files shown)


Dean177:
Thanks @tchak

PR #147, opened 2020-06-15
feat(string): add indexOf and indexOfRight

### LINKED ISSUE #146: Feature Request: String.get, String.indexOf
Hey all, thanks for the hard work on Tablecloth (and Standard) and for deciding to merge the two together, I think it's the right move even though compromises like that can be very tough!

Would the two following operations make sense to add? No preference for naming.

`String.get` exists in `Standard` which is handy since the Stdlib-version throws.
`String.indexOf` exists in `Stdlib` as `index_opt`

EDIT: Semi-related and I realise this may be personal preference, but I also kind of miss utility-functions like `String.isEmpty(foo)` vs doing `foo == ""`

lessp:
I don't know which convention you'd like to follow (I know JS uses `indexOf` and `lastIndexOf`) or if it's idiomatic to use `~char` as a named argument. 🙂 

Not too familiar with the OCaml-syntax, so there may be some traces of `refmt` left in there! Anyway, let me know how you'd like to proceed!

Would close #146

### DIFF
--- a/bucklescript/test/StringTest.re
+++ b/bucklescript/test/StringTest.re
@@ -28,47 +28,41 @@ let suite =
       });
     });
 
-    describe("firstIndexOf", () => {
-      test("returns some index of the first matching character", () => {
-        expect(
-          firstIndexOf("hello", ~char='h') |> Option.unwrap(~default=9999),
-        )
-        |> toEqual(Eq.int, 0)
+    describe("indexOf", () => {
+      test("returns some index of the first matching substring", () => {
+        expect(indexOf("hello", "h"))
+        |> toEqual(Eq.(option(int)), Some(0))
       });
 
       test("returns the first index even though multiple present", () => {
-        expect(
-          firstIndexOf("hellh", ~char='h') |> Option.unwrap(~default=9999),
-        )
-        |> toEqual(Eq.int, 0)
+        expect(indexOf("hellh", "h"))
+        |> toEqual(Eq.(option(int)), Some(0))
       });
 
-      test("returns None when no character matches", () => {
-        expect(
-          firstIndexOf("hello", ~char='x') |> Option.unwrap(~default=0),
-        )
-        |> toEqual(Eq.int, 0)
+      test("returns first substring that matches with multiple characters", () => {
+        expect(indexOf("hellh", "ell"))
+        |> toEqual(Eq.(option(int)), Some(1))
+      });
+
+      test("returns None when no substring matches", () => {
+        expect(indexOf("hello", "xy")) |> toEqual(Eq.(option(int)), None)
       });
     });
 
-    describe("lastIndexOf", () => {
-      test("returns some index of the last matching character", () => {
-        expect(
-          lastIndexOf("hello", ~char='o') |> Option.unwrap(~default=9999),
-        )
-        |> toEqual(Eq.int, 4)
+    describe("indexOfRight", () => {
+      test("returns some index of the last matching string", () => {
+        expect(indexOfRight("helloh", "oh"))
+        |> toEqual(Eq.(option(int)), Some(4))
       });
 
       test("returns the last index even though multiple present", () => {
-        expect(
-          lastIndexOf("hellh", ~char='h') |> Option.unwrap(~default=9999),
-        )
-        |> toEqual(Eq.int, 4)
+        expect(indexOfRight("ohelloh", "oh"))
+        |> toEqual(Eq.(option(int)), Some(5))
       });
 
       test("returns None when no character matches", () => {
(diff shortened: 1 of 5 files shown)


Dean177:
Sorry to quickly change my mind on this but I think we should have a slightly different API:

`findIndex: string -> ~f:(char -> bool) -> (int * char) option` which mirrors the functions available in the Array & List modules 

`indexOf : string -> string -> int option`

Which you use like: `String.indexOf "Hello World World" "World" = Some 6` and the corresponding `indexOfRight`.

Don't worry about adding `findIndex` that can be a separate PR

Thanks for the help!

bucklescript/test/StringTest.re
```diff
@@ -28,6 +28,50 @@ let suite =
       });
     });
 
+    describe("firstIndexOf", () => {
+      test("returns some index of the first matching character", () => {
+        expect(
+          firstIndexOf("hello", ~char='h') |> Option.unwrap(~default=9999),
+        )
+        |> toEqual(Eq.int, 0)
```
Dean177, line 36: There is an Eq.option function which is used like so:
```
|> toEqual(Eq.(option(int)), Some(0))
```

That means you won't need the `Option.unwrap` in your tests.

lessp:
> Sorry to quickly change my mind on this but I think we should have a slightly different API:
> 
> `findIndex: string -> ~f:(char -> bool) -> (int * char) option` which mirrors the functions available in the Array & List modules
> 
> `indexOf : string -> string -> int option`
> 
> Which you use like: `String.indexOf "Hello World World" "World" = Some 6` and the corresponding `indexOfRight`.
> 
> Don't worry about adding `findIndex` that can be a separate PR
> 
> Thanks for the help!

I'd still do `lastIndexOf` as well though, right? 

One could add `indexOfChar` too I guess, but probably better as a separate discussion! 🙂

Dean177:
> ~I'd still do `lastIndexOf` as well though, right?~

Yep.

> To me `lastIndexOf` is a bit more declarative than `indexOfRight` (and perhaps more align to other languages, Java, JavaScript) but might be easier to find `indexOfRight` next to `indexOf`. 

Thats part of the idea, the other part is we already have functions like `foldRight`, `trimRight`, `padRight` etc

> One could add `indexOfChar` too I guess, but probably better as a separate discussion! 🙂

My thinking was that `findIndex` would cover the use case for `indexOfChar` like so:

```
let indexOfChar string character = 
  findIndex string ~f:(fun _index element -> Char.equal element character)
```

lessp:
> > ~I'd still do `lastIndexOf` as well though, right?~
> 
> Yep.
> 
> > To me `lastIndexOf` is a bit more declarative than `indexOfRight` (and perhaps more align to other languages, Java, JavaScript) but might be easier to find `indexOfRight` next to `indexOf`.
> 
> Thats part of the idea, the other part is we already have functions like `foldRight`, `trimRight`, `padRight` so the idea of `Right` is pretty well established as a convention.
> 
> > One could add `indexOfChar` too I guess, but probably better as a separate discussion! 🙂
> 
> My thinking was that `findIndex` would cover the use case for `indexOfChar` like so:
> 
> ```
> let indexOfChar string character = 
>   findIndex string ~f:(fun _index element -> Char.equal element character)
> ```

Cool, makes sense! I'll get around to it later! 👍

Dean177:
I added findIndex as a separate issue in #156 so it isn't forgotten

lessp:
I was unable to get the builds from the Makefile working:

```
❯ make build

Building tablecloth-native ...
opam config exec -- dune build
File "native/src/TableclothOption.ml", line 5, characters 13-27:
Error: Unbound module Option
File "native/src/TableclothResult.ml", line 8, characters 52-61:
Error: Unbound module Result
File "native/src/TableclothArray.ml", line 68, characters 44-52:
Error: Unbound module Fun
File "native/src/TableclothList.ml", line 126, characters 43-51:
Error: Unbound module Fun
File "native/src/TableclothMap.ml", line 55, characters 38-48:
Error: Unbound module Option
make[1]: *** [build-native] Error 1
make: *** [build] Error 2
```

Dean177:
Which version of ocaml are you using?

You may need to create a switch for at least 4.08.

bucklescript/test/StringTest.re
lessp: Switched in: d497d57

Dean177:
Awesome, thanks @lessp.

All you need to do to fix the `documentation-generator-build` is run `
cd ocamldoc-json-generator && make doc` then commit the updated model.json.

lessp:
> Awesome, thanks @lessp.
> 
> All you need to do to fix the `documentation-generator-build` is run ` cd ocamldoc-json-generator && make doc` then commit the updated model.json.

Sweet, looks like it's all passing now!

PR #164, opened 2020-06-22
Upgrade the minimum version of base

### LINKED ISSUE #163: Native doesn't build with base v0.14.0
Trying to tun the project show me this error: 
Screen Shot 2020-06-21 at 1 21 12 PM

So to run the project you should downgrade the base to v0.13.2 with this 
`opam pin add base v0.13.2`

This would downgrade the base and being able to run it

Dean177:
Resolves #163

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test:
 deps-native:
 	@printf "\n\e[31mInstalling native dependencies ...\e[0m\n"
 	opam update
-	opam install alcotest base dune junit junit_alcotest odoc reason -y
+	opam install alcotest "base>=v0.14.0" dune junit junit_alcotest odoc reason -y
 	@printf "\n\e[31mInstalled!\e[0m\n"
 
 deps-bs:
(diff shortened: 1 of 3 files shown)


pbiggar:
OK, so base requires ocaml >= 4.07, which means you can't use the same version of OCaml for bs and native. So I think we shouldn't do this. Is there another way to solve this problem?

I guess it's time to build matrixes to circleci, to make sure we have all our bases (heh) covered.

Dean177:
The other thing we could do is set an upper limit for the version of base.

I don't think the native portion of the codebase builds with a version less than 4.08 as is since the native implementation relies on the `Result` & `Option` modules.

What are the downsides to requiring different minimum compiler versions?

pbiggar:
If you use both different versions, Merlin won't work (seems impossible to
fix, though would love to be wrong on this, see
https://reasonml.chat/t/ocaml-4-06-1-and-reasonml-frontend/2364)

pbiggar:
This is handled in #181

Issue #175, opened 2020-09-14
Make build gives me an error.


shubhamkumar13:
I followed the developer setup guide -> https://github.com/darklang/tablecloth#developing.

And I got an error in `native/src/TableclothMap.ml` 

```
sk@pop-os:~/tablecloth (master)$ make build
make[1]: Entering directory '/home/sk/tablecloth'

Building tablecloth-native ...
opam config exec -- dune build
File "native/src/TableclothMap.ml", line 100, characters 20-29:
100 |           | true -> `Fst data
                          ^^^^^^^^^
Error: This expression has type [> `Fst of 'a ]
       but an expression was expected of type ('b, 'c) Base.Either.t
Done: 436/455 (jobs: 1)make[1]: *** [Makefile:3: build-native] Error 1
make[1]: Leaving directory '/home/sk/tablecloth'
make: *** [Makefile:12: build] Error 2
```

Looking at the source code made me realize that nothing is wrong. This function is dependent on `Base.Map.partition_mapi` which emits a tuple -> http://binaryanalysisplatform.github.io/bap/api/odoc/base/Base/Map/index.html#val-partition_mapi

So I cannot see what's going on here, any suggestions would be helpful :)

pbiggar:
It looks like newer version of base use an `Either` instead of <code>`Fst</code> and <code>`Snd</code>. See https://ocaml.janestreet.com/ocaml-core/v0.14/doc/base/Base/Map/ vs https://ocaml.janestreet.com/ocaml-core/v0.10/doc/base/Base/Map/.

Not sure how to fix that in a general solution that will work with all versions. It may be that we should just update to the format of the latest version?

pbiggar:
This is fixed in #220

PR #176, opened 2020-09-14
orElse for TableclothResult

shubhamkumar13:
Refer #162 
Option already has an [`orElse`](https://github.com/darklang/tablecloth/blob/eca87dac1f2c3742526b67983b9da8a23f4bf3ba/bucklescript/src/TableclothOption.ml)

This PR tries to replicate the same functionality for Result's `orElse`

PR #177, opened 2020-09-16
Update types present in TableclothMap.ml's Base.Either from `Fst / `Snd to First/Second

shubhamkumar13:
Updated types in response to issue #175

### DIFF
--- a/native/src/TableclothMap.ml
+++ b/native/src/TableclothMap.ml
@@ -96,7 +96,7 @@ let filter = Base.Map.filter
 
 let partition m ~f =
   Base.Map.partition_mapi m ~f:(fun ~key ~data ->
-      if f ~key ~value:data then `Fst data else `Snd data)
+      if f ~key ~value:data then First data else Second data)
 
 
 let find m ~f =


shubhamkumar13:
@pbiggar I think #176 can be merged after this is merged and I rebase with the new master

pbiggar:
Great, thanks! Where are those types from, and what versions will this work on? I'm worried about it working for new versions and breaking for old versions.

Sorry for the hassle with these changes, really appreciate them. I'm trying to think what would provide certainty here - I think adding a build matrix to CircleCI with the different versions of Base would certainly do it - is that something you can do?

Wouldn't mind some input from @Dean177 here too.

shubhamkumar13:
> and what versions will this work on?

I'm currently using `ocaml-base-compiler.4.10.0`

> Where are those types from

These types are from [native/src/TableclothMap.ml](https://github.com/darklang/tablecloth/blob/83480ad7b58c24fae96c246712d4082f2ff610e1/native/src/TableclothMap.ml#L99)

So as you stated in #175 `partition_mapi` from `Base.Map` uses `Base.Either` now instead of using ``[> `Fst of x | `Snd of x ]``
> I think adding a build matrix to CircleCI with the different versions of Base would certainly do it - is that something you can do?

I have not used CircleCI but I can try looking into it

pbiggar:
I added CircleCI and a list of supported versions. I think either this isn't necessary anymore, or instead that supporting this will add extra supported versions (which would be great). Can you rebase this off main and we can see what CI says?

pbiggar:
This is being fixed in #220. I managed to avoid using `\`Fst`/`First` and reusing an existing function instead

Issue #179, opened 2020-09-21
Support multiple versions of Base/OCaml/Bucklescript


pbiggar:
There are several major versions of ocaml, base, and bucklescript in the wild. Clearly, if we are going to wrap them we need some kind of story for how to handle this.

This is affecting users and contributors: #176, #177, #163, #141, #175.

One option is to only support the latest of each, allowing older versions of the library to be used for older upstream versions. I think considering this has a slower update frequency than some of the upstream libraries, this isn't necessarily great. Also, many people use these old versions, myself included.

To support multiple versions, we would need to:
- [x] add CircleCI build matrix support to test on multiple versions
- [x] move dependencies to be build dependencies
- [x] provide accurate ranges of what we support in opam
- [x] update README with what we support

PR #181, opened 2020-09-22
Add build matrix for native on CircleCI

pbiggar:
It seems that the master branch works on native for OCaml 4.08, 4.09 and 4.10, using base v0.12 and v013. I couldn't test 4.11, and other versions of base did not work for reasons I didn't investigate.

It seems like this is a good starting point to document what we support and to start to support other versions.

Fixes half of https://github.com/darklang/tablecloth/issues/179


pbiggar:
Going to merge this and #189 to unblock contributors.

Issue #217, opened 2021-07-30
Map.String is not clonable


pbiggar:
When attempting to send a Map.String.t to a web worker, I got an error:

```
Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': function caml_compare(_a, _b) {
  while(true) {
    var b = _b;
    var a = _a;
    if (a === b) {
      return...<omitted>...
} could not be cloned.
```

This was not the case in 0.0.7. I think the difference is that StrDict in 0.0.7 was a [Belt.MapString](https://rescript-lang.org/docs/manual/latest/api/belt/map-string), which had a built-in comparator.

pbiggar:
I investigated this more. The situation is that Belt.Map has a cmp field in the generated object, which holds a function. That function (like all functions) can't be serialized using the cloning algorithm used by web workers.

I tried to fix this in https://github.com/darklang/tablecloth/pull/229. This PR separated Set.String.t from Set.t, making the former a Belt.Set.String, which has a built-in comparator and does not suffer from the cmp field issue.

Ultimately, this became very annoying very fast. It had a number of issues:
- all the Set functions needed to be copied to Set.String and Set.Int (I couldn't find any way to reuse Belt.Set or its functions, and so needed to make a whole new set of functionality)
- the interface would no longer be the same as the native version of Set.String
- My code needs to change to use Tablecloth.Map.String.* everywhere, versus at the moment you can do Tablcloth.Map.* for nearly all functions. These were hundreds of changes
- Any time I wanted to create a Map that wasn't of Int or Strings, that initial (unclonable) problem continued to exist.

I

pbiggar:
This will be solved by #285, leaving open to confirm.

PR #229, opened 2021-08-02
Removes cmp field from Set.{String,Int}.t

pbiggar:
See discussion in https://github.com/darklang/tablecloth/issues/217.


pbiggar:
Closing as this will be solved a different way I think

PR #231, opened 2021-08-16
Missing functions

kate-grant:
Added List.sortBy and List.sort_by to TableclothList.ml and TableclothList.mli


pbiggar:
Looks good, thanks!

Needs an ocaml version as well, which you should be able to find in the history.

You should rebase off main to remove the changelog commits too.

pbiggar:
Looks great, thanks!

PR #233, opened 2021-10-30
Move Rescript to camlCase and Ocaml to snake_case

Lomand:
Removed snacke_case functions and two tests that were comparing interfaces with native ocaml counterparts.


Lomand:
Native builds will fail untill separate `.mli` files will not be created for them, I'll address this in another PR request (with ocaml camlCase functions removal).

pbiggar:
This is great, thank you!

I'd like to keep things green - can you copy over the old .mlis to the native directory before we merge this so it's green (don't need to do the ocaml changes in the same PR)?

I'm trying to see we should organize this, given that there's a release that needs to go out without these changes. Perhaps we should keep a `next` branch?

Lomand:
Hmm... Yep, moving to `next` branch seems reasonable since you want to release current version without this changes. Also as it turns out moving ocaml to snake_case and rescript to camelCase in a single pull request makes much more sense, I'll update the PR accordingly.

Lomand:
Since `native` and `rescript` are now completely decoupled, they can no longer be represented fairly at '/api' webpage. For that reason I've modified ocamldoc generator to create separate versions of 'model.json', one for native and another for rescript version of the project. Making ocamldoc and rescript-compiler work together was a major pain and ultimately it is done by generating entire documentation with 'melange.re', so no incompatible `.cmi` files and other Ocaml 4.06 shenanigans. There are still some problems with generation of the rescript model, the generator may throw
```
File "./_rescript/TableclothFloat.ml", line 79, characters 24-25:
79 | external largestValue : t = "MAX_VALUE" [@@bs.scope "Number"] [@@bs.val]
                             ^
Error: External identifiers must be functions
```
for no apparent reason since `largestValue` is successfully added to `model-rescript.json`, same thing with `Warning: Element List not found`, List is in `model-rescript.json` and successfully is displayed on the `/api` page. What is puzzling is that the native `model.json` is now missing some functions. Since it does not produce any errors I am not sure why that happens, will check on that later. I'll also try adapting the documentation page for new multi-implementation format, some very basic steps are already done.

API - Google Chrome 2021-11-07 19-19-16

pbiggar (commented):
This looks great, thank you! One question and we should be ready to merge.

pbiggar:
(Apologies for the slow review aswell!)

Lomand:
No worries, we're not in a rush 😉 . I'll add a makefile and tidy up some minor things (like moving `esy` to `PATH`) on the weekend. I'll also add `rescript` syntax alongside `reason` and `ocaml` in `model.json` and adjust `/api` page in another PR.

Thanks for creating tablecloth, loving it.

pbiggar:
Would it be worth replacing the reason with the rescript one? It seems to me that reason is functionally dead, right?

Lomand:
Well, the only downside of keeping reason examples is enlarged filesize of the `model.json`. You are correct though, there is zero to none traction in reason native comunity, essentially it is dead.

pbiggar:
Yeah, that's what I thought. Probably better to drop support then and focus on rescript.

Lomand:
I've moved the code into the makefile rule just as you asked. Also, I've double checked `model.json` for any missing content, and turns out I've accidentally deleted about 10 comments, they are restored now. Removed filtering on `api` page code, there is no need to hide functions with `_` in their name. I think the PR should be ready to merge.

pbiggar:
Wonderful, thank you!

PR #234, opened 2021-11-11
Add missing tests

kate-grant:
added tests, added conditionals for negative numbers to List.range and List.repeat, rewrote List.map2, List.map3, and List.zip to handle lists of different lengths, added List.groupi and rewrote List.groupWhile with groupi, changed init index in Array.splitAt, reversed return of Array.values and Array.filtermap

### DIFF
--- a/rescript/test/OptionTest.ml
+++ b/rescript/test/OptionTest.ml
@@ -277,4 +277,16 @@ let suite =
               expect (Option.equal Int.equal (Some 1) None)
               |> toEqual Eq.bool false ) ;
           test "returns bool true if options are equal" (fun () ->
-              expect (Option.equal Int.equal None None) |> toEqual Eq.bool true ) ) )
+              expect (Option.equal Int.equal None None) |> toEqual Eq.bool true ) ) ;
+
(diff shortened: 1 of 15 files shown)


rescript/test/OptionTest.ml
```diff
@@ -9,4 +9,284 @@ let suite =
...
+                   None ) ) ;
+
+      describe "unwrap" (fun () ->
+          test "returns unwrapped [option('a)]" (fun () ->
+              expect (Option.unwrap ~default:99 (Some 42)) |> toEqual Eq.int 42 ) ;
+
+          test "returns default" (fun () ->
+              expect (Option.unwrap ~default:99 None) |> toEqual Eq.int 99 ) ;
+
+          test "returns default" (fun () ->
```
pbiggar, line 216: Unnecessary test I think

pbiggar (changes requested):
This is great! There's a couple of small changes and we should be able to merge then :)

pbiggar:
This is great, thanks.

Discussed changes in person. I'll merge this now - we can do the change to the descriptions in another PR.

Thanks again!

PR #235, opened 2021-12-24
API Page UX Improvements

Lomand:
From the user's perspective, there are some things that the current `/api` page lacks in terms of readability. 
I've tried to address that and make documentation more readable and on par with lodash/rambda docs.

In this PR I'll also plan to implement a switcher between F#/Ocaml/Rescript versions.

Current version:
image

This PR:
image

Since the docs are generated from the source files, I'll need to remove empty lines in the examples section. I still need to add mobile styles as well as tweaks to the dark theme to make this PR ready to merge.


pbiggar:
This is a big improvement!

Can I suggest you keep the `let length : 'a Array.t -> int` instead of just going down to `'a Array.t -> int`. It seems more readable, and also looks quite nice in the old formatting (vs dropping to the basic `code` format).

Lomand:
Sure thing, I'll address that.

Lomand:
I've added the mobile theme and night mode support with some additional changes. To make Tablecloth more familiar to the average developer I've made the following:

The urls of some pages were changed to more reflect the actual content and resemble the counterparts of similar projects (Lodash & Rambda):

### `/documentation/installation` -> `/get-started/installation`
This page with other introduction materials were already labeled as "Get started" on the home page even before this PR. Now the urls will be representative with the actual content. TLDR: thouse pages live under `get-started` prefix.


### `/api` -> `/docs/rescript` `/docs/ocaml` `/docs/reason` `/docs/fsharp`
Two changes here: now each of the languages gets its own url, so examples can be linked with preffered language/syntax already preselected (ex: `/docs/rescript#Bool.xor` and `/docs/ocaml#Float.subtract`), the url itself was changed from `api` to `docs` since inlining urls to other projects will provide more plesent UX for the average developer, who is already knows what's in the docs page since it used Lodash before.

pbiggar:
This looks phenomenal! I love the logos in particular and also the url scheme is great.

Do we need reason anymore? AIUI, we don't have reason code anymore.

Lomand:
Thanks for your feedback! I'll remove Reason from the docs page.

Lomand:
Looks like it's done:
* Removed Reason from the docs
* Fixed generator of header id's in the `get-started` pages (ex:`/installation#ocaml-native#using-opam` -> `/installation#using-opam` and `#___unsafe-means-"could-raise-an-exception"` -> `___unsafe-means-could-raise-an-exception`)

Cleaned up code, it is now ready to be merged in `next`.

pbiggar:
Wonderful! Will give this a read this weekend.

Separately, we've started to release version 0.0.8, so we should be able to merge the next branch soon!

PR #238, opened 2022-01-05
Add Result.orElse & fix #162

### LINKED ISSUE #162: Add Result.orElse and Option.orElse
These should work similarly to the corresponding `or_` functions but with the arguments swapped:

```ocaml
Ok "the value I want"
|> Result.orElse (Ok "a default value")
= Ok “the value I want”
```

pm5:


pbiggar:
Wonderful, thank you.

I've got a bit of an ordering problem here as I still haven't reviewed #234, so let me get on that ASAP and then I'll come back to this.

pbiggar:
This looks great, thank you!

Looks like the documentation needs to be rebuilt - I may do that today to get this merged.

pbiggar:
Merged in another PR. Thanks so much for the contribution!

PR #239, opened 2022-01-09
Type information in Rescript syntax in Docs

Lomand:
Rescript part of the Docs should have examples and type definitions displayed in Rescript syntax (Right now both are displayed in Ocaml). 

Example: `'a -> 'b -> 'c -> 'a * 'b * 'c` -> `('a, 'b, 'c) => ('a, 'b, 'c)`
People with no experience in Ocaml will find the first version confusing and cryptic.

Since currently there is no way to use Rescript parser programmatically (https://github.com/melange-re/melange/issues/228) the only option is to convert interfaces from `.mli` to `.resi`, and read the type information from them. Sound suboptimal, but works surprisingly well (488/499 types converted without a hitch). There were some problems (https://github.com/rescript-lang/syntax/pull/449) with `Tablecloth.Map.String` and `Tablecloth.Set.String` type definition (`type nonrec 'value t = 'value Of(TableclothString).t`), I had to rewrite the type to equivalent 
```
  type identity

  type nonrec 'value t = (TableclothString.t, 'value, identity) t
```
Since you are in process of refactoring the current implementation (https://github.com/darklang/tablecloth/pull/229) this will no longer be an issue when this PR merges.

There is a need for some testing just to check if everything is alright and also fix broken type definitions in submodules.

I also want to convert all examples in rescript docs to rescript syntax, I'll address this in another PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,7 @@ jobs:
       - run: npm install -g esy
       - run: cd ocamldoc-json-generator && make init-node
       - run: cd ocamldoc-json-generator && ~/.npm-global/bin/esy
+      - run: cd ocamldoc-json-generator && make rescript-interfaces
       - run: cd ocamldoc-json-generator && make doc
       # Check that model.json has been regenerated when there are changes in the interface files
       - run: git diff --exit-code website/model.json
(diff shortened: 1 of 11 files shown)


Lomand:
The PR is ready. The only type that failed to generate is `Internal.toBeltComparator`. It's not exposed externally in the library, so this issue can be ignored (this type is not displayed in the docs). Other than that, everything is looking normal.

pbiggar:
great, thanks! I looked at it over the weekend, will try to get it merged today.

pbiggar:
This is great, thank you!

PR #240, opened 2022-01-10
Add Result.unwrapLazy & closed #159

### LINKED ISSUE #159: Add Result.unwrapLazy
This should work the same way as Result.unwrap but the default value should be a 'a lazy rather than a 'a
```ocaml
val getLazy : 'a Result. -> default:('a Lazy.t) -> 'a
```

pm5:


pbiggar:
This looks useful, thanks!

I think this is the first thing we're adding that uses lazy. Any thoughts on using `lazy` vs using a function (eg `Result.unwrapLazy r ~default:(fun () -> Ok 5`)?

pbiggar:
Needs formatting and the doc to rebuild. There's a list of commands to do that here: https://github.com/darklang/tablecloth#developing

pm5:
> Needs formatting and the doc to rebuild. There's a list of commands to do that here: https://github.com/darklang/tablecloth#developing

That's nice :D

pm5:
> This looks useful, thanks!
> 
> I think this is the first thing we're adding that uses lazy. Any thoughts on using `lazy` vs using a function (eg `Result.unwrapLazy r ~default:(fun () -> Ok 5`)?

I prefer `lazy` with it's monadic operators most of the time.  Don't actually know if there's any impact on performance though.

PR #241, opened 2022-01-10
updated path prefix

kate-grant:
added static asset path prefix to gatsby config


pbiggar:
Were you able to get it fully deployed with this?

I think it would be better to merge this as part of making the CI do the full deploy, as it won't be useful otherwise.

PR #296, opened 2022-10-12
Support F#

OceanOak:


pbiggar (changes requested):
Great, thanks!

A couple of small fixes before we'd want to merge:
- rename directory to "fsharp"
- remove the `bin` directory. When removing it, can you make sure it's not in the git history either.

